PR #1, opened 2019-09-09
Updating readme to trigger tests

maxhutch:


scripts/strip_hints.sh
```diff
@@ -0,0 +1,12 @@
...
+find ./src -regex '.*\.py'|while read fileName; do
+  echo $fileName
+  strip-hints $fileName --to-empty --only-assigns-and-defs >tmp.py
+  mv tmp.py $fileName
+done
+find ./tests -regex '.*\.py'|while read fileName; do
+  echo $fileName
+  strip-hints $fileName --to-empty --only-assigns-and-defs >tmp.py
+  mv tmp.py $fileName
+done
```
bfolie: This second block can be removed, since all of the files that might have typehints are in the taurus/ directory, as opposed to being spread across a src/ and a tests/ directory.

PR #7, opened 2019-09-12
PLA-1288 Fix incomplete rehydration of data objects

asantas93:
Rather than try to substitute links with objects in the json object
hook, simply populate the link -> object index and then substitute
once the index is fully populated.


asantas93:
I tested this by running `test_material_history` 25 times against development with my local taurus changes. In each test, I inspected `cake_history.process.ingredients[1].spec`, the object that sometimes did not get replaced properly, according to PLA-1288, and each time it was an actual spec object.

asantas93:
The aforementioned test may not have been sufficient. Please hold off on review until I do more testing.

asantas93:
As usual, my taurus package installation was screwed up, but I seem to have fixed it now. Test I added fails before this change and now passes. Ready for review.

PR #13, opened 2019-09-18
Feature/serialized json material history

kroenlein:
Merge in the history serialization.

Code restructure to mimic process is fairly orthogonal.


kroenlein:
If this isn't ready for prime time quite yet, I'll get on this since I need the working code to generate corpus for Ranveer & Danny.

PR #14, opened 2019-09-18
Numpy-style docstrings for miscellaneous entity classes

bfolie:


taurus/entity/setters.py
```diff
@@ -13,7 +31,20 @@ def validate_list(obj, typ, trigger=None):
...
+
+    Parameters
+    ----------
+    obj: Any
+        Object that should be a string
+
+    Returns
+    -------
+    str
+        Return obj if it is a string, otherwise throw ValueError
```
kroenlein: Is this the right way to document potential exceptions in this formalism?

bfolie: Good question. Turns out they do have a `Raises` section, which I have now included.

PR #16, opened 2019-09-19
We should be doing all content-based validation server-side

maxhutch:
This removes content based validation, but leaves in place
type based validation.  The motivation here is that the presense
of LinkByUIDs makes validation of un-liked objects incomplete
and therefore partially inconsistent.  Validation should be the
responsibility of the data store, not the data objects themselves.

At Citrine, this means that we implement content-based validation
server side, which let's us include validations that traverse links
between the different data objects.  If we support client-side validation,
it will be through making a call to our data service.


maxhutch:
Historical note: we implemented a lot of this content-based validation in `taurus` as a prototype for logic that has since been moved into Citrine's data service.  This is more of a `move` than a `remove`.

taurus/entity/template/process_template.py
```diff
@@ -51,26 +51,3 @@ def allowed_labels(self, allowed_labels):
...
-            extra = unique_names.difference(set(self.allowed_names))
-            if len(extra) > 0:
-                raise ValueError("Found disallowed names: {}".format(extra))
-
-        # Check the label namespace
-        if self.allowed_labels is not None:
-            labels = {label for x in process.ingredients for label in x.labels}
-            extra = labels.difference(set(self.allowed_labels))
-            if len(extra) > 0:
-                raise ValueError("Found disallowed labels: {}".format(extra))
```
bfolie: I do not disagree with removing this

taurus/entity/template/base_template.py
```diff
@@ -36,35 +31,3 @@ def _homogenize_ranges(template_or_tuple):
...
-
-                # if the attribute doesn't have a template, assign this template to it
-                if attribute.template is None:
-                    attribute.template = attr_template
-                # if it does have a template and its not a LinkByUID, make sure its the right one
-                elif attribute.template != attr_template \
-                        and isinstance(attribute.template, AttributeTemplate) \
-                        and isinstance(attr_template, AttributeTemplate):
-                    raise ValueError(
-                        "Template and object templates don't match for all {}".format(attr_name))
```
bfolie: I do not disagree with removing this

taurus/entity/object/has_quantities.py
```diff
@@ -69,12 +78,6 @@ def absolute_quantity(self, absolute_quantity):
         if absolute_quantity is None:
             self._absolute_quantity = None
         elif isinstance(absolute_quantity, ContinuousValue):
-            if isinstance(absolute_quantity, NominalReal) and absolute_quantity.nominal < 0:
-                raise ValueError("absolute_quantity must be positive")
-            elif isinstance(absolute_quantity, NormalReal) and absolute_quantity.mean < 0:
-                raise ValueError("absolute_quantity must be positive")
-            elif isinstance(absolute_quantity, UniformReal) and absolute_quantity.lower_bound < 0:
-                raise ValueError("absolute_quantity must be positive")
```
bfolie: I disagree with removing this

bfolie:
It seems reasonable to remove all validation that may or may not be do-able because it involves resources such as templates where the local client may only have a LinkByUID or it may have a resource that is out of date. Furthermore, template validation seems like it is an additional constraint we are imposing that is not inherent to the data model, just like a Taurus user is free to create a measurement run that has no `material` field, but that object would not be accepted by our database.

It does _not_ make sense to me to remove validation on objects that do not have UIDs. For example I see no reason to not do validation on a RealBounds. In any use case of this library, a RealBounds with lower_bound > upper_bound will always be invalid, and should throw an error.

taurus/entity/bounds/tests/test_real_bounds.py
```diff
@@ -73,9 +43,4 @@ def test_constructor_error():
 def test_type_mismatch():
     """Test that incompatible types cannot be matched against RealBounds."""
     bounds = RealBounds(0, 1, default_units="meters")
-    assert not bounds.validate(NominalCategorical("foo"))
     assert not bounds.contains(IntegerBounds(0, 1))
-    with pytest.raises(TypeError):
-        bounds.validate(0)
-    with pytest.raises(TypeError):
-        bounds.contains([.33, .66])
```
bfolie: I disagree with removing this

taurus/entity/bounds/tests/test_integer_bounds.py
```diff
@@ -21,24 +18,9 @@ def test_incompatible_types():
...
-
-    assert int_bounds.validate(NominalInteger(0))
-    assert int_bounds.validate(NominalInteger(2))
-    assert not int_bounds.validate(NominalInteger(3))
-
-    assert not int_bounds.validate(UniformInteger(0, 3))
-    with pytest.raises(TypeError):
-        int_bounds.validate(0)
-
-
```
bfolie: I disagree with removing this

taurus/entity/bounds/tests/test_composition_bounds.py
```diff
@@ -25,16 +23,6 @@ def test_invalid_constructor():
...
-def test_validate():
-    """Test basic validation logic."""
-    bounds = CompositionBounds(components={"spam", "eggs"})
-    assert bounds.validate(NominalComposition({"spam": 1.0, "eggs": 2.0}))
-    assert not bounds.validate(NominalComposition({"foo": 1.0, "bar": 2.0}))
-    assert not bounds.validate(NominalReal(2.0, ''))
-    with pytest.raises(TypeError):
-        bounds.validate({"spam": 1.0, "eggs": 2.0})
-
-
```
bfolie: I disagree with removing this

taurus/entity/bounds/tests/test_categorical_bounds.py
```diff
@@ -25,16 +23,6 @@ def test_invalid_constructor():
...
-def test_validate():
-    """Test basic validation logic."""
-    bounds = CategoricalBounds(categories={"spam", "eggs"})
-    assert bounds.validate(NominalCategorical("spam"))
-    assert not bounds.validate(NominalCategorical("foo"))
-    assert not bounds.validate(NominalReal(2.0, ''))
-    with pytest.raises(TypeError):
-        bounds.validate("spam")
-
-
```
bfolie: I disagree with removing this

taurus/entity/bounds/real_bounds.py
```diff
@@ -51,44 +46,6 @@ def default_units(self, default_units):
...
-            return upper >= value.nominal >= lower
-
-        # Normal distributions are interpreted as truncated normals,
-        # so long as their median value is within the bounds
-        if isinstance(value, NormalReal):
-            return upper >= value.mean >= lower
-
-        if isinstance(value, UniformReal):
-            return upper >= value.upper_bound and lower <= value.lower_bound
-
```
bfolie: I disagree with removing this

taurus/entity/bounds/integer_bounds.py
```diff
@@ -31,34 +27,6 @@ def __init__(self, lower_bound=None, upper_bound=None):
...
-            return False
-        if not isinstance(value, IntegerValue):
-            return False
-
-        if isinstance(value, NominalInteger):
-            return self.upper_bound >= value.nominal >= self.lower_bound
-
-        if isinstance(value, UniformInteger):
-            return self.upper_bound >= value.upper_bound and self.lower_bound <= value.lower_bound
-
```
bfolie: I disagree with removing this

taurus/entity/bounds/composition_bounds.py
```diff
@@ -41,30 +39,6 @@ def components(self, value):
...
-            True if the value is a composition with all required components.
-
-        """
-        if not super().validate(value):
-            return False
-        if not isinstance(value, CompositionValue):
-            return False
-
-        return all(x in self.components for x in value.components)
-
```
bfolie: I disagree with removing this

taurus/entity/bounds/categorical_bounds.py
```diff
@@ -43,37 +39,6 @@ def categories(self, categories):
...
-
-        if isinstance(value, DiscreteCategorical):
-            return all(x in self.categories for x in value.probabilities)
-
-        if isinstance(value, NominalCategorical):
-            return value.category in self.categories
-
-        msg = "Categorical bounds do not check every CategoricalValue subclass"  # pragma: no cover
-        assert False, msg  # pragma: no cover
-
```
bfolie: I disagree with removing this

taurus/entity/bounds/base_bounds.py
```diff
@@ -2,34 +2,11 @@
...
-        bool
-            True if the value is a member of the bounds, and False otherwise
-
-        """
-        if value is None:
-            return False
-        if isinstance(value, BaseValue):
-            return True
-        raise TypeError('{} is not a Value object'.format(value))
-
```
bfolie: I disagree with removing this

taurus/entity/attribute/base_attribute.py
```diff
@@ -60,13 +53,7 @@ def template(self):
     def template(self, template):
         if template is None:
             self._template = None
-        elif isinstance(template, LinkByUID):
-            self._template = template
-        elif isinstance(template, AttributeTemplate):
-            if not template.validate(self):
-                raise ValueError("Template is incompatible with the attribute")
-            if self.value and not template.bounds.validate(self.value):
-                raise ValueError("the template is inconsistent with the value")
```
bfolie: I do not disagree with removing this

taurus/entity/bounds/tests/test_real_bounds.py
maxhutch: This leaves `contains`, so I should have left this check for a type error on `contains`; nice catch!

taurus/entity/bounds/tests/test_integer_bounds.py
maxhutch: This isn't changing the test coverage; its a consequence of removing the `validate` method.

taurus/entity/bounds/tests/test_composition_bounds.py
maxhutch: This isn't changing the test coverage; its a consequence of removing the `validate` method.

taurus/entity/bounds/tests/test_categorical_bounds.py
maxhutch: This isn't changing the test coverage; its a consequence of removing the `validate` method.

taurus/entity/bounds/real_bounds.py
maxhutch: But you're OK with removing the other cases?  I agree that we should add the requirement that upper >= lower in the bounds constructor.  A bit part of of this removal is meant to provide consistency of functionality across cases.

taurus/entity/bounds/integer_bounds.py
maxhutch: See comment in `RealBounds`

taurus/entity/bounds/composition_bounds.py
maxhutch: If we don't have validate on all bounds, then we shouldn't have it on any.

bfolie: I think we should validate all values against bounds. I only support removing validation when it touches templates or other resources that have an external life someplace else.

PR #20, opened 2019-09-22
Add PerformedSource

maxhutch:
Per https://github.com/CitrineInformatics/taurus-documentation/blob/master/docs/specification/objects.md#source

This is an optional addition, so non-breaking.

### DIFF
--- a/taurus/entity/object/has_source.py
+++ b/taurus/entity/object/has_source.py
@@ -3,7 +3,7 @@ from taurus.entity.source.performed_source import PerformedSource
 
 
 class HasSource(object):
-    """Mixin-trait for entities that include parameters."""
+    """Mixin-trait for entities that include sources (data provenance)."""
 
     def __init__(self, source):
         self._source = None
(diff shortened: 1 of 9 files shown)


taurus/entity/object/has_source.py
```diff
@@ -0,0 +1,24 @@
+"""For entities that have parameters."""
+from taurus.entity.source.performed_source import PerformedSource
+
+
+class HasSource(object):
+    """Mixin-trait for entities that include parameters."""
```
kroenlein:
```suggestion
    """Mixin-trait for entities that include sources (data provenance)."""
```

maxhutch:
> Look good. I presume we'll be checking string formatting on the backend.

Yes, but we don't yet.  I went ahead and added parsing/validation/formatting using the same library that we use in `citrine-python`.

kroenlein:
Is this consistent with what we decided around parsing formulas?

maxhutch:
> Is this consistent with what we decided around parsing formulas?

I think so, because I think ISO 8601 is well defined enough to constitute a "type" whereas the payloads that we want to support for "empirical formula" are not.  As an analog, I'd consider InChI/SMILES to be "types" (even though I know there are some subtle difference in their parsers).  If we level-up the specification such that we actually have a well-defined "empirical formula" type, then I'd advocate that we go ahead and implement a python parser for it and do that client side.  I just don't think we're that specific yet.

PR #21, opened 2019-09-23
Down-select the units supported in pint.

maxhutch:
Pint supports _many_ units.  If we want to maintain the ability
to use different unit parsing libraries, then we should down select
the set of supported units to make the task of finding equivalents
easier.

This list can easily be added to as needed, of course.

### DIFF
--- a/taurus/entity/bounds/tests/test_real_bounds.py
+++ b/taurus/entity/bounds/tests/test_real_bounds.py
@@ -31,8 +31,8 @@ def test_boundsless():
 
 def test_non_prefix():
     """Make sure validation works when the units differ by more than a prefix."""
-    dim2 = RealBounds(lower_bound=0, upper_bound=100, default_units="amp")
-    assert dim2.validate(NominalReal(50, "V / ohm"))
+    dim2 = RealBounds(lower_bound=0, upper_bound=100, default_units="N")
+    assert dim2.validate(NominalReal(50, "kg m / sec^2"))
 
 
 def test_contains():
--- a/taurus/ingest/material_run_example.py
+++ b/taurus/ingest/material_run_example.py
@@ -20,7 +20,7 @@ known_properties = {
     ),
     "kinematic viscosity": PropertyTemplate(
         name="kinematic viscosity",
-        bounds=RealBounds(lower_bound=0.0, upper_bound=10.0**40, default_units="m^2/s")
+        bounds=RealBounds(lower_bound=0.0, upper_bound=10.0**40, default_units="")
     )
 }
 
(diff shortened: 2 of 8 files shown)


taurus/entity/bounds/tests/test_real_bounds.py
```diff
@@ -31,8 +31,8 @@ def test_boundsless():
 
 def test_non_prefix():
     """Make sure validation works when the units differ by more than a prefix."""
-    dim2 = RealBounds(lower_bound=0, upper_bound=100, default_units="amp")
```
maxhutch: We don't have to support charge-related units right now

taurus/ingest/tests/test_material_run_example.py
```diff
@@ -35,7 +35,7 @@
             "knob_2_setting": "high",
             "temperature": 624,
             "density": "0.80 +- 0.12",
-            "kinematic viscosity": "1.0e3 St",
```
maxhutch: We don't have to support Stokes right now

taurus/ingest/material_run_example.py
```diff
@@ -20,7 +20,7 @@
     ),
     "kinematic viscosity": PropertyTemplate(
         name="kinematic viscosity",
-        bounds=RealBounds(lower_bound=0.0, upper_bound=10.0**40, default_units="m^2/s")
+        bounds=RealBounds(lower_bound=0.0, upper_bound=10.0**40, default_units="")
```
kroenlein: Why is this non-dimensional?  Meter and second are supported.

taurus/ingest/tests/test_material_run_example.py
```diff
@@ -35,7 +35,7 @@
...
+            "kinematic viscosity": "0.1",
```
kroenlein: Why remove viscosity units?  This would seem important to polymer and melt studies.

taurus/ingest/material_run_example.py
maxhutch: Because the parsing logic that I'm using in this example is terrible and gets stuck on spaces.  I can just cram the units together though.  Update coming.

taurus/ingest/tests/test_material_run_example.py
maxhutch: Its not on @lkubie 's list.

taurus/ingest/material_run_example.py
kroenlein: Do we need the PR in before the update is ready?

maxhutch: No; I appended to this PR.

taurus/ingest/material_run_example.py
```diff
@@ -16,18 +16,18 @@
 known_properties = {
     "density": PropertyTemplate(
         name="density",
-        bounds=RealBounds(lower_bound=0.0, upper_bound=1000.0, default_units='')
+        bounds=RealBounds(lower_bound=0.0, upper_bound=1000.0, default_units='g / cm^3')
```
kroenlein: For reference, `pint` recognizes `cc`

PR #26, opened 2019-09-23
Feature/pla 1413 repr has soft links

bfolie:
When displaying the `__repr__()` of a `DictSerializable`, include soft links. Those soft-linked objects have their pointers replaced with LinkByUID, to prevent infinite recursion loops.
I had to change the soft-linked `ValidList` objects (`measurements` in `MaterialRun`, `ingredients` in `ProcessRun` and `ProcessSpec`) to allow `LinkByUID`.

### DIFF
--- a/taurus/entity/dict_serializable.py
+++ b/taurus/entity/dict_serializable.py
@@ -82,7 +82,11 @@ class DictSerializable(ABC):
         return loads(dumps(d))
 
     def __repr__(self):
-        return str(self.as_dict())
+        object_dict = self.as_dict()
+        skipped_keys = {x.lstrip('_') for x in vars(self) if x in self.skip}
+        for key in skipped_keys:
+            object_dict[key] = self.__getattribute__(key)
+        return str(object_dict)
 
     def __eq__(self, other):
         if isinstance(other, DictSerializable):
--- a/taurus/entity/dict_serializable.py
+++ b/taurus/entity/dict_serializable.py
@@ -1,6 +1,5 @@
 from abc import ABC
 import json
-from copy import deepcopy
 
 
 class DictSerializable(ABC):
@@ -83,21 +82,40 @@ class DictSerializable(ABC):
         return loads(dumps(d))
 
     def __repr__(self):
-        from taurus.util.impl import substitute_links, set_uuids
         object_dict = self.as_dict()
-        # as_dict skips over keys in `skip`, but they should be in the representation.
+        # as_dict() skips over keys in `skip`, but they should be in the representation.
         skipped_keys = {x.lstrip('_') for x in vars(self) if x in self.skip}
         for key in skipped_keys:
-            skipped_attribute = deepcopy(self.__getattribute__(key))
-            # Replace links in skipped keys with LinkByUID to prevent infinite recursion loop.
-            set_uuids(skipped_attribute)
-            if isinstance(skipped_attribute, list):
-                for x in skipped_attribute:
-                    substitute_links(x)
-            else:
-                substitute_links(skipped_attribute)
-            object_dict[key] = skipped_attribute
+            skipped_field = self.__getattribute__(key)
+            object_dict[key] = self._name_repr(skipped_field)
         return object_dict.__str__()
+
+    def _name_repr(self, entity):
+        """
+        A representation of an object or a list of objects that uses the name and type.
+
+        This is used to represent soft-linked objects without inundating the user with
+        repetitive information.
+
+        Parameters
+        ----------
+        entity: DictSerializable or List[DictSerializable]
+            Object to represent using its name. Generally a (list of) BaseEntity or BaseAttribute,
+            both of which have a `name` field.
+
+        Returns
+        -------
+        str
+            A representation of `entity` using its name.
+
+        """
+        if isinstance(entity, (list, tuple)):
+            return [self._name_repr(item) for item in entity]
+        elif entity is None:
+            return None
+        else:
+            name = getattr(entity, 'name', '<unknown name>')
+            return "<{} '{}'>".format(type(entity).__name__, name)
 
     def __eq__(self, other):
         if isinstance(other, DictSerializable):
--- a/taurus/entity/dict_serializable.py
+++ b/taurus/entity/dict_serializable.py
@@ -86,9 +86,9 @@ class DictSerializable(ABC):
         # as_dict() skips over keys in `skip`, but they should be in the representation.
         skipped_keys = {x.lstrip('_') for x in vars(self) if x in self.skip}
         for key in skipped_keys:
-            skipped_field = self.__getattribute__(key)
+            skipped_field = getattr(self, key, None)
             object_dict[key] = self._name_repr(skipped_field)
-        return object_dict.__str__()
+        return str(object_dict)
 
     def _name_repr(self, entity):
         """
--- a/taurus/entity/dict_serializable.py
+++ b/taurus/entity/dict_serializable.py
@@ -1,5 +1,6 @@
 from abc import ABC
 import json
+from copy import deepcopy
 
 
 class DictSerializable(ABC):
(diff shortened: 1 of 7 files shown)


taurus/entity/dict_serializable.py
```diff
@@ -82,7 +83,21 @@ def build(d):
...
-        return str(self.as_dict())
+        from taurus.util.impl import substitute_links, set_uuids
+        object_dict = self.as_dict()
+        # as_dict skips over keys in `skip`, but they should be in the representation.
+        skipped_keys = {x.lstrip('_') for x in vars(self) if x in self.skip}
+        for key in skipped_keys:
+            skipped_attribute = deepcopy(self.__getattribute__(key))
+            # Replace links in skipped keys with LinkByUID to prevent infinite recursion loop.
+            set_uuids(skipped_attribute)
+            if isinstance(skipped_attribute, list):
```
bfolie: Without this if statement, the `measurements` field in a `MaterialRun` will be a list of `LinkByUID`.

taurus/entity/dict_serializable.py
```diff
@@ -82,7 +83,21 @@ def build(d):
...
     def __repr__(self):
-        return str(self.as_dict())
+        from taurus.util.impl import substitute_links, set_uuids
+        object_dict = self.as_dict()
+        # as_dict skips over keys in `skip`, but they should be in the representation.
+        skipped_keys = {x.lstrip('_') for x in vars(self) if x in self.skip}
+        for key in skipped_keys:
+            skipped_attribute = deepcopy(self.__getattribute__(key))
+            # Replace links in skipped keys with LinkByUID to prevent infinite recursion loop.
+            set_uuids(skipped_attribute)
```
asantas93: I'm a little worried about a mutating an object in a display function like this, even if it's on a deep copy of an object. Since this is a "soft link" anyway, can you just make the skipped items show up as links instead of replacing their children with links? Open to other suggestions, too.

asantas93: I guess there's no guarantee that the children will have an id yet. I will think about this some more.

maxhutch: If we're setting uuids here in the copy, then they won't actually point back to the original object, right?

bfolie: The impetus here is that when a user displays an object in a Jupyter notebook they see `__repr__()`. And the idea of a "soft link" is not relevant to the user. From their point of view the material run has measurements. So if they see that the `measurements` field in the material run is missing (what happens now), or they just see a list of LinkByUID, it gives them the sense that the measurements are not there/have been turned into links. That's why I wanted to make the actual object appear in `__repr__`.

bfolie: > I guess there's no guarantee that the children will have an id yet. I will think about this some more.
In Citrine-Python we automatically generate IDs as soon as every object is created, so this will never be invoked. But in Taurus there's no guarantee that the object has uids.

asantas93: Well the original context was just that people thought they were literally just missing. I think the user will realize they can just print out `material.measurements` as long as `measurements` should up at all in the output. We don't need to recreate the whole tree here, they can dump JSON if they need that.

asantas93: I think `measurements: <List[MeasurementRun]>` would probably be sufficient to convince the user that their data is there.

asantas93: Seems like if you called this twice, you'd see different ids each time.

asantas93: Well I don't think the copy would point to the wrong objects (or at least not the wrong ids), but if you made a link to the generated id here because you saw it in the output of a print statement, it seems like it would be an invalid one.

asantas93: The copy only lives for the lifetime of the `__repr__` function afaict

bfolie: The uid in this case is meaningless, because the actual object doesn't know about it.

bfolie: We could make it a real action and actually set the uids.

bfolie: That's reasonable, but I'd like to run it by strike-team-data before committing to anything.

taurus/entity/dict_serializable.py
```diff
@@ -82,7 +82,40 @@ def build(d):
         return loads(dumps(d))
 
     def __repr__(self):
-        return str(self.as_dict())
+        object_dict = self.as_dict()
+        # as_dict() skips over keys in `skip`, but they should be in the representation.
+        skipped_keys = {x.lstrip('_') for x in vars(self) if x in self.skip}
+        for key in skipped_keys:
+            skipped_field = self.__getattribute__(key)
```
asantas93: Typically I think you would use the equivalent `getattr(self, key)` syntax for this.

taurus/entity/dict_serializable.py
```diff
@@ -82,7 +82,40 @@ def build(d):
...
+            object_dict[key] = self._name_repr(skipped_field)
+        return object_dict.__str__()
```
asantas93: Similarly, you can use `str(object_dict)` -- in general any built-in function defined like `__foo__(self)` can be invoked with `foo(object_implementing___foo__)`

taurus/entity/dict_serializable.py
```diff
@@ -82,7 +82,40 @@ def build(d):
...
+        -------
+        str
+            A representation of `entity` using its name.
+
+        """
+        if isinstance(entity, (list, tuple)):
+            return [self._name_repr(item) for item in entity]
+        elif entity is None:
+            return None
+        else:
```
asantas93: I think it would be a good idea to just fall back on `repr(entity)` if `entity` isn't actually an instance of `DictSerializable`. I know that won't come up right now but it seems possible that in the future we may want to put a logger or some other unserializable object in `_skip` which violates the assumption about type in this function.
Alternatively, you could do that check in `__repr__` itself.

taurus/entity/object/tests/test_material_run.py
```diff
@@ -59,6 +59,9 @@ def test_process_run():
     copy_material = loads(dumps(material_run))
     assert dumps(copy_material) == dumps(material_run)
 
+    assert 'output_material' in process_run.__repr__()
```
asantas93: You could write this as `repr(process_run)`.

taurus/entity/dict_serializable.py
bfolie: Done

taurus/entity/dict_serializable.py
bfolie: Done

taurus/entity/object/tests/test_material_run.py
bfolie: Done

taurus/entity/dict_serializable.py
bfolie: I had something like that initially, but removed it because there was no non-artificial way to test it. There's no way to make `__repr__` call `_name_repr` on an object that isn't a `DictSerializable`, because it only gets called for soft links and soft links cannot be set directly. Therefore in order to get coverage on that line I would have had to make a new test that created a non-DictSerializable object and called `_name_repr` on it, a displeasingly artificial situation.
Note that there are actually no type assumptions here. If we pass an object of type `Logger` that has no `name` field, the representation would be `"<Logger '<unknown name>'>"`. It is easy to in the future modify this if-else block to take other options into account, but since we want 100% coverage I did not want to preemptively check for situations that are presently incapable of occurring.

bfolie:
> Some style and minor design nits, feel free to merge if you have more important stuff to do now.

I am glad to include style improvements. One of your comments I have a reason for not implementing and that is explained in a comment.

taurus/entity/dict_serializable.py
asantas93: Oh I see, yeah I'm not worried about this at all really then, thanks for explaining!

PR #29, opened 2019-09-24
Increase test coverage with substitute_links() and ignoring demos

bfolie:
Write tests to cover every line of `taurus.util.impl.substitute_links()` and update config file so that `demo/` is ignored. Bumps coverage to 94%.

### DIFF
--- a/taurus/util/tests/test_substitute_links.py
+++ b/taurus/util/tests/test_substitute_links.py
@@ -1,9 +1,10 @@
 """Test the substitute_links method, in particular the edge cases that the client doesn't test."""
 import pytest
+from uuid import uuid4
 
 from taurus.util.impl import substitute_links
-from taurus.entity.object.measurement_run import MeasurementRun
-from taurus.entity.object.material_run import MaterialRun
+from taurus.entity.link_by_uid import LinkByUID
+from taurus.entity.object import MeasurementRun, MaterialRun, ProcessRun, ProcessSpec
 
 
 def test_substitution_without_id():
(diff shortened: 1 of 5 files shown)


taurus/util/tests/test_substitute_links.py
```diff
@@ -0,0 +1,67 @@
...
+    meas = MeasurementRun("A measurement", material=mat,
+                          uids={native_id: str(uuid4()), other_id: str(uuid4())})
+
+    # Turn the material pointer into a LinkByUID using native_id
+    substitute_links(meas, native_uid=native_id)
+    assert meas.material == LinkByUID.from_entity(mat, name=native_id)
+
+    # Put the measurement into a list and convert that into a LinkByUID using native_id
+    measurements_list = [meas]
+    substitute_links(measurements_list)
```
maxhutch:
```suggestion
    substitute_links(measurements_list, native_uid=native_id)
```

taurus/util/tests/test_substitute_links.py
```diff
@@ -0,0 +1,67 @@
...
+    with pytest.raises(ValueError):
+        substitute_links({mat: meas}), "substitute_links should fail if objects don't have uids"
+    with pytest.raises(ValueError):
+        substitute_links({meas: mat}), "substitute_links should fail if objects don't have uids"
+
+
+def test_native_id_substitution():
+    """Test that the native id gets serialized, when specified."""
+    native_id = 'id1'
+    other_id = 'id2'
```
maxhutch: You might want to add more other_ids so the probability of randomly selecting the `native_id` is lower.

taurus/util/tests/test_substitute_links.py
bfolie: Good catch. Fixed

taurus/util/tests/test_substitute_links.py
bfolie: Done

PR #31, opened 2019-09-25
Full test coverage of objects

bfolie:
Complete test coverage of `taurus/entity/object/` directory. Bumps coverage to 95.7%

### DIFF
--- a/taurus/entity/object/tests/test_ingredient_spec.py
+++ b/taurus/entity/object/tests/test_ingredient_spec.py
@@ -74,3 +74,11 @@ def test_invalid_quantities(invalid_quantity):
         IngredientSpec(name="name", number_fraction=invalid_quantity)
     with pytest.raises(TypeError):
         IngredientSpec(name="name", absolute_quantity=invalid_quantity)
+
+
+def test_invalid_assignment():
+    """Invalid assignments to `process` or `material` throw a ValueError."""
+    with pytest.raises(ValueError):
+        IngredientSpec(name="name", material=NominalReal(3, ''))
+    with pytest.raises(ValueError):
+        IngredientSpec(name="name", process="process")
--- a/taurus/entity/object/tests/test_ingredient_spec.py
+++ b/taurus/entity/object/tests/test_ingredient_spec.py
@@ -79,8 +79,8 @@ def test_invalid_quantities(invalid_quantity):
 
 
 def test_invalid_assignment():
-    """Invalid assignments to `process` or `material` throw a ValueError."""
-    with pytest.raises(ValueError):
+    """Invalid assignments to `process` or `material` throw a TypeError."""
+    with pytest.raises(TypeError):
         IngredientSpec(name="name", material=NominalReal(3, ''))
-    with pytest.raises(ValueError):
+    with pytest.raises(TypeError):
         IngredientSpec(name="name", process="process")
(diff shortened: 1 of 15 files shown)


taurus/entity/object/tests/test_ingredient_spec.py
```diff
@@ -22,3 +29,56 @@ def test_ingredient_reassignment():
...
+@pytest.mark.parametrize("valid_quantity", VALID_QUANTITIES)
+def test_valid_quantities(valid_quantity):
+    """
+    Check that all quantities must be continuous values.
+
+    There are no restrictions on the value or the units. Although a volume fraction of -5 kg
+    does not make physical sense, it will not throw an error.
+    """
+    ingred = IngredientSpec(name="name", mass_fraction=valid_quantity)
+    assert ingred.mass_fraction == valid_quantity
```
kroenlein: Shouldn't any mass fraction with a dimensional unit be an error?  Or does that qualify as a ValueError instead of a TypeError? (BE responsibility)

taurus/entity/object/tests/test_ingredient_spec.py
```diff
@@ -22,3 +29,56 @@ def test_ingredient_reassignment():
...
+    ingred = IngredientSpec(name="name", volume_fraction=valid_quantity)
+    assert ingred.volume_fraction == valid_quantity
```
kroenlein: As above

taurus/entity/object/tests/test_ingredient_spec.py
```diff
@@ -22,3 +29,56 @@ def test_ingredient_reassignment():
...
+    ingred = IngredientSpec(name="name", number_fraction=valid_quantity)
+    assert ingred.number_fraction == valid_quantity
```
kroenlein: As above

taurus/entity/object/tests/test_ingredient_spec.py
```diff
@@ -22,3 +29,56 @@ def test_ingredient_reassignment():
...
+    NominalReal(14.0, ''),
+    UniformReal(0.5, 0.6, 'm'),
+    NormalReal(-0.3, 0.6, "kg")
+]
+
+INVALID_QUANTITIES = [
+    NominalCategorical("blue"),
+    NominalInteger(5),
+    0.33,
+    "0.5"
```
kroenlein: You may consider throwing an EmpiricalFormula in there, too

taurus/entity/object/tests/test_process_run.py
```diff
@@ -35,3 +40,20 @@ def test_ingredient_run():
 
     assert 'process' in repr(ingred_run)
     assert 'ingredients' in repr(proc_run)
+
+
+def test_invalid_assignment():
+    """Invalid assignments to `spec` throw a ValueError."""
+    with pytest.raises(ValueError):
+        ProcessRun("name", spec=[ProcessSpec("spec")])
```
kroenlein: Shouldn't this and the similar exceptions be TypeErrors not ValueErrors?

taurus/entity/object/tests/test_ingredient_spec.py
bfolie: We've removed that validation from Taurus and put it in the backend. And since Taurus accepts any ContinuousValue, even those that are physically nonsense, I decided to make the tests reflect that (this is explained in the test's docstring)

taurus/entity/object/tests/test_ingredient_spec.py
bfolie: Will do

taurus/entity/object/tests/test_process_run.py
bfolie: That's a good point. They should be TypeErrors. I suspect there are quite a few places where that should be changed. I'll do that.

PR #32, opened 2019-09-25
Test coverage of values

bfolie:
Full coverage of modules in `taurus/entity/values/`. This bumps coverage to 96.6%. Also change some ValueErrors to TypeErrors, fix bug in `categorical` assignment of `NominalCategorical`, and check that `UniformInteger` bounds are ints.

### DIFF
--- a/taurus/entity/value/tests/test_uniform_real.py
+++ b/taurus/entity/value/tests/test_uniform_real.py
@@ -21,3 +21,4 @@ def test_equality():
     assert value1 == value1
     assert value1 != value2  # Equality check does not do unit conversion.
     assert value1 != value3
+    assert value1 != 0.5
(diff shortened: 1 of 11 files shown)


taurus/entity/value/tests/test_uniform_real.py
```diff
@@ -0,0 +1,24 @@
...
+    UniformReal(4.4, 8.8, 'm')
+    UniformReal(100.0, 100.0, 'm')
+    with pytest.raises(AssertionError):
+        UniformReal(23.2, 18.9, 'm')
+
+
+def test_equality():
+    """Test that equality checks both bounds and integers."""
+    value1 = UniformReal(0, 1, '')
+    value2 = UniformReal(0, 100, 'cm')
```
kroenlein: Do we want to check pint functionality?

bfolie: Fair. I updated it to be a simple check of the units strings (which __eq__ does compare).

PR #34, opened 2019-09-26
Test coverage of `taurus/entities` directory

bfolie:
Tests of bounds, `build`, equality, instantiating uids using a list,  `allowed_names`, `allowed_labels`, `content_hash`, `make_instance`, and the `ValidList` class.

Brings coverage to 98.3%

### DIFF
--- a/taurus/entity/bounds/real_bounds.py
+++ b/taurus/entity/bounds/real_bounds.py
@@ -93,19 +93,11 @@ class RealBounds(BaseBounds):
             A tuple of the (lower_bound, upper_bound) in the target units.
 
         """
-        # if neither have units, then just return the bounds
-        if not self.default_units and not target_units:
-            return self.lower_bound, self.upper_bound
-        # if either have units but both don't, then we can't return anything
-        elif not self.default_units or not target_units:
+        try:
+            lower_bound = units.convert_units(
+                self.lower_bound, self.default_units, target_units)
+            upper_bound = units.convert_units(
+                self.upper_bound, self.default_units, target_units)
+            return lower_bound, upper_bound
+        except units.IncompatibleUnitsError:
             return None, None
-        # if both have units, then we can try to convert
-        else:
-            try:
-                lower_bound = units.convert_units(
-                    self.lower_bound, self.default_units, target_units)
-                upper_bound = units.convert_units(
-                    self.upper_bound, self.default_units, target_units)
-                return lower_bound, upper_bound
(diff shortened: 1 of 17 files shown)


taurus/entity/bounds/real_bounds.py
```diff
@@ -93,19 +93,11 @@ def _convert_bounds(self, target_units):
             A tuple of the (lower_bound, upper_bound) in the target units.
 
         """
-        # if neither have units, then just return the bounds
-        if not self.default_units and not target_units:
-            return self.lower_bound, self.upper_bound
-        # if either have units but both don't, then we can't return anything
-        elif not self.default_units or not target_units:
```
bfolie: RealBounds are required to have units now, so there is no need to have these checks.

taurus/entity/util.py
```diff
@@ -84,9 +84,9 @@ def array_like():
         try:
             import pandas as pd
             _array_like = (list, tuple, np.ndarray, pd.core.base.PandasObject)
-        except ImportError:
-            _array_like = (list, tuple, np.ndarray)
-    except ImportError:
-        _array_like = (list, tuple)
+        except ImportError:  # pragma: no cover
```
bfolie: This is hard to test because it only occurs if the local environment doesn't have numpy and/or pandas installed.

taurus/entity/tests/test_entity.py
```diff
@@ -12,3 +16,17 @@ def test_id_case_sensitivitiy():
     ingredient = IngredientRun(uids={'my_id': 'sample1'})
     assert ingredient.uids['my_id'] == 'sample1'
     assert ingredient.uids['MY_id'] == 'sample1'
+
+
+def test_content_hash():
```
bfolie: I'm not sure what the content_hash is to be used for, so I don't know what about it to test.

taurus/entity/template/base_template.py
```diff
@@ -66,4 +66,4 @@ def _homogenize_ranges(template_or_tuple):
                     return [first, second]
                 else:
                     raise ValueError("Range and template are inconsistent")
-        raise ValueError("Expected a template or (template, bounds) tuple")
+        raise TypeError("Expected a template or (template, bounds) tuple")  # pragma: no cover
```
bfolie: This line cannot be covered naturally because the methods that call `_homogenize_ranges` already check that the argument they are passing is an `AttributeTemplate`, `list`, or `tuple`

PR #41, opened 2019-10-03
Cake ingredient names

kroenlein:
Added names to cake.py example so that Graph story has a compliant object

### DIFF
--- a/taurus/demo/cake.py
+++ b/taurus/demo/cake.py
@@ -32,6 +32,7 @@ from taurus.entity.util import complete_material_history
 
 def make_cake():
     """Define all objects that go into making a demo cake."""
+    
     cake = MaterialRun(name='Cake')
 
     baked = MaterialRun(name='Baked Cake')
--- a/taurus/demo/cake.py
+++ b/taurus/demo/cake.py
@@ -55,34 +55,44 @@ def make_cake():
 
     cake.process = ProcessRun(name='Icing')
     for ingredient in (baked, frosting):
-        IngredientRun(material=ingredient, process=cake.process)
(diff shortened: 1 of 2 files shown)


taurus/demo/cake.py
```diff
@@ -126,27 +136,38 @@ def make_cake():
...
+                                  material=run.material.spec, process=drymix.process.spec)
 
     set_uuids(cake)
 
-    IngredientSpec(material=LinkByUID.from_entity(butter.spec), process=frosting.spec.process)
-    IngredientSpec(material=chocolate.spec, process=frosting.spec.process)
-    IngredientSpec(material=powder_sugar.spec, process=frosting.spec.process)
-    IngredientSpec(material=LinkByUID.from_entity(vanilla.spec), process=frosting.spec.process)
-    IngredientSpec(material=LinkByUID.from_entity(milk.spec), process=frosting.spec.process)
+    print(butter)
```
bfolie: debugging statement?

taurus/demo/cake.py
```diff
@@ -126,27 +136,38 @@ def make_cake():
...
+
+    for run in frosting.process.ingredients:
```
bfolie: Suggest changing `run` to `ingredient_run`, for readability

taurus/demo/cake.py
kroenlein: Yes, good catch.

taurus/demo/cake.py
kroenlein: Changing to ing_run as compromise, given flake8 line length constraint.

PR #54, opened 2019-11-19
Add complete set of value types to cake example

kroenlein:
Built out a comprehensive set of values so that cake is a comprehensive representation of the GEMD model.

* Added PropertyAndConditions (Nutritional Information, Expected Taste, Empirical Formulas)
* Added Categoricals (Nutritional Information, Doneness metrics)
* Added Compositions (Empirical Formulas, Nutritional Information)
* Changed a Nominal Integer to a Uniform Integer (measured taste)
* Modified method calls to allow for template, spec permanence across executions

TODO:
* Figure out how to persist Measurement Specs

### DIFF
--- a/taurus/demo/cake.py
+++ b/taurus/demo/cake.py
@@ -83,9 +83,9 @@ def make_cake_templates():
     tmpl["Nutritional Information"] = PropertyTemplate(
         name="Nutritional Information",
         description="FDA Nutrition Facts, mass basis.  Please be attentive to g vs. mg.  "
-                    + "`other-carbohydrate` and `other-fat` are the total values minus the "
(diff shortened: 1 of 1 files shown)


taurus/demo/cake.py
```diff
@@ -306,11 +401,14 @@ def make_cake_spec():
             tags=[
                 'purchase::dry-goods'
             ],
-            notes='Purchasing salt'
+            notes='Purchasing salt',
```
bfolie: Why is this comma here?

kroenlein: Good catch, I think, thought it's not clear why flake didn't catch this
```suggestion
            notes='Purchasing salt'
```

PR #56, opened 2019-11-27
Feature/pla 1951 citrine python taurus integration

crleblanc:
This PR has two small commits.  One fixes a bug regarding mutating dictionary keys while iterating over it.  This operation raises an exception in newer versions of Python such as 3.8.0 and could output incorrect data.

The other commit uses thin_dumps when serializing DictSerializable objects.  This is required in order to use these objects with citrine-python, e.g. registering a Taurus ProcessSpec in a Dataset.  I suggest we create an integration test to make sure it's properly covered.

Please let me know if switching to thin_dumps might cause any unintended consequences.

This is based on the PR from @asantas93.

### DIFF
--- a/taurus/client/json_encoder.py
+++ b/taurus/client/json_encoder.py
@@ -140,8 +140,8 @@ def thin_dumps(obj, **kwargs):
 
     Parameters
     ----------
-    obj: BaseEntity
-        Object to dump (must be a BaseEntity because it must have uids)
+    obj:
+        Object to dump
     **kwargs: keyword args, optional
         Optional keyword arguments to pass to `json.dumps()`.
 
@@ -151,8 +151,6 @@ def thin_dumps(obj, **kwargs):
         A serialized string of `obj`, with link_by_uid in place of pointers to other objects.
 
     """
-    if not isinstance(obj, BaseEntity):
-        raise TypeError("Can only dump BaseEntities, but got {}".format(type(obj)))
 
     set_uuids(obj)
     res = deepcopy(obj)
--- a/taurus/client/json_encoder.py
+++ b/taurus/client/json_encoder.py
@@ -151,7 +151,6 @@ def thin_dumps(obj, **kwargs):
         A serialized string of `obj`, with link_by_uid in place of pointers to other objects.
 
     """
-
     set_uuids(obj)
     res = deepcopy(obj)
     substitute_links(res)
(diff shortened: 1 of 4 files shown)


crleblanc:
@mcgalcode it looks like I'm unable to add reviewers to this repo.

@maxhutch and @asantas93, it would be great if you could look at the PR and see if there are any issues.

taurus/util/impl.py
```diff
@@ -65,20 +65,22 @@ def substitute(thing):
                 else:
                     substitute(x)
         elif isinstance(thing, dict):
-            for k, v in thing.items():
+            # we are restricted from modifying the keys of a dict while iterating over it's
```
kroenlein:
```suggestion
            # we are restricted from modifying the keys of a dict while iterating over its
```

taurus/client/tests/test_json.py
```diff
@@ -78,8 +78,9 @@ def test_thin_dumps():
     assert isinstance(thin_copy.spec, LinkByUID)
     assert thin_copy.spec.id == meas_spec.uids['my_scope']
 
-    with pytest.raises(TypeError):
-        thin_dumps(LinkByUID('scope', 'id'))
+    # Check that LinkByUID objects are correctly converted their JSON equivalent
+    expected_json = '{"id": "my_id", "scope": "scope", "type": "link_by_uid"}'
+    assert thin_dumps(LinkByUID('scope', 'my_id')) == expected_json
```
kroenlein: Does thin_dumps enforce/provide key ordering?  This seems like it could cause spurious test failures if the keys weren't output alphabetically.

taurus/client/json_encoder.py
```diff
@@ -140,8 +140,8 @@ def thin_dumps(obj, **kwargs):
 
     Parameters
     ----------
-    obj: BaseEntity
-        Object to dump (must be a BaseEntity because it must have uids)
+    obj:
+        Object to dump
```
kroenlein: Do we have a test that verifies good behavior when we're passed a non-BaseEntity?

taurus/client/tests/test_json.py
crleblanc: That's a good question, I'll look into this.

crleblanc: I can see that thin_dumps uses sort_keys=True when using json.dumps.  These are the same parameters as used by the dumps function in json_encoder.py.  I think it's safe to assume that key ordering is unchanged.

taurus/client/json_encoder.py
crleblanc: I added a test for this using a dict with a string key and value.  This raises a TypeError which seems appropriate.  I'll update the PR with this check.

PR #57, opened 2019-12-11
Bug in recursive foreach: it doesn't walk from process to ingredients

maxhutch:
This is just a reproduction step.  Fix to come.

Fix:
The seen set of object hashes should be enough to avoid cycles,
so we don't have to re-use the skip member from DictSerializable
(which conflated serialization and pointer concerns).

Also moved the applyFirst/applyLast logic out of those if statements
to prevent the first object in a cycle from having the function applied
to it twice.

### DIFF
--- a/taurus/entity/util.py
+++ b/taurus/entity/util.py
@@ -1,4 +1,5 @@
 """Utility methods."""
+from taurus.util import recursive_foreach
 
 
 def make_instance(base_spec):
@@ -104,31 +105,17 @@ def complete_material_history(mat):
         links substituted.
     """
     from taurus.entity.base_entity import BaseEntity
-    from taurus.entity.dict_serializable import DictSerializable
     import json
     from taurus.client.json_encoder import dumps, loads
     from taurus.util.impl import substitute_links
 
-    queue = [mat]
-    seen = set()
     result = []
 
-    while queue:
-        obj = queue.pop(0)
+    def body(obj: BaseEntity):
+        copy = loads(dumps(obj))
+        substitute_links(copy)
+        result.append(json.loads(dumps(copy))[0][0])
 
-        if isinstance(obj, BaseEntity):
-            if obj not in seen:
-                seen.add(obj)
-                queue.extend(obj.__dict__.values())
-
-                copy = loads(dumps(obj))
-                substitute_links(copy)
-                result.insert(0, json.loads(dumps(copy))[0][0])  # Leaf first
-        elif isinstance(obj, (list, tuple)):
-            queue.extend(obj)
-        elif isinstance(obj, dict):
-            queue.extend(obj.values())
-        elif isinstance(obj, DictSerializable):
-            queue.extend(obj.__dict__.values())
+    recursive_foreach(mat, body, apply_first=False)
 
     return result
--- a/taurus/util/impl.py
+++ b/taurus/util/impl.py
@@ -153,41 +153,40 @@ def recursive_foreach(obj, func, apply_first=False, seen=None):
         else:
             seen.add(obj)
 
+    if apply_first and isinstance(obj, BaseEntity):
+        func(obj)
+
     if isinstance(obj, (list, tuple)):
         for i, x in enumerate(obj):
             if isinstance(x, BaseEntity):
                 if apply_first:
-                    func(x)
                     recursive_foreach(x, func, apply_first, seen)
                 else:
                     recursive_foreach(x, func, apply_first, seen)
-                    func(x)
             else:
                 recursive_foreach(x, func, apply_first, seen)
     elif isinstance(obj, dict):
         for x in concatv(obj.keys(), obj.values()):
             if isinstance(x, BaseEntity):
                 if apply_first:
-                    func(x)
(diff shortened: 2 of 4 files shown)


maxhutch:
Note that I was trying to do a TDD pattern with the test, but I forgot to `git add` and then didn't wait to see the build go red.  I did run the test locally, and it did fail.

taurus/util/impl.py
```diff
@@ -153,41 +153,40 @@ def recursive_foreach(obj, func, apply_first=False, seen=None):
         else:
             seen.add(obj)
 
+    if apply_first and isinstance(obj, BaseEntity):
+        func(obj)
+
     if isinstance(obj, (list, tuple)):
         for i, x in enumerate(obj):
             if isinstance(x, BaseEntity):
                 if apply_first:
```
kroenlein: This is now a no-op

taurus/util/tests/test_foreach.py
```diff
@@ -0,0 +1,15 @@
+from taurus.entity.object import ProcessRun, MaterialRun, IngredientRun
+from taurus.util.impl import recursive_foreach
+
+
+def test_recursive_foreach():
```
kroenlein: This test needs to do a full process->material->ingredient->process->material set in order to actually guarantee behavior.

maxhutch: Why do you say that?  If there's something to add, I'd guess that it should be a measurement rather than another process.

taurus/util/impl.py
```diff
@@ -153,41 +153,40 @@ def recursive_foreach(obj, func, apply_first=False, seen=None):
...
                 else:
                     recursive_foreach(x, func, apply_first, seen)
-                    func(x)
             else:
                 recursive_foreach(x, func, apply_first, seen)
     elif isinstance(obj, dict):
         for x in concatv(obj.keys(), obj.values()):
             if isinstance(x, BaseEntity):
                 if apply_first:
-                    func(x)
```
bfolie: We no longer want to apply func(x) on dict or DictSerializable that isn't a BaseEntity?

taurus/util/tests/test_foreach.py
bfolie: Agreed (both that an additional process->material link is not necessary, and that a measurement should be added)

taurus/util/impl.py
maxhutch: We never did.

taurus/util/tests/test_foreach.py
maxhutch: Done, plus a property template for good measure.

taurus/util/impl.py
maxhutch: :+1: simplified

taurus/entity/util.py
```diff
@@ -104,31 +105,17 @@ def complete_material_history(mat):
...
-                copy = loads(dumps(obj))
-                substitute_links(copy)
-                result.insert(0, json.loads(dumps(copy))[0][0])  # Leaf first
-        elif isinstance(obj, (list, tuple)):
-            queue.extend(obj)
-        elif isinstance(obj, dict):
-            queue.extend(obj.values())
-        elif isinstance(obj, DictSerializable):
-            queue.extend(obj.__dict__.values())
+    def body(obj: BaseEntity):
```
maxhutch: @kroenlein : I went ahead and applied it here for good measure

maxhutch:
> In addtion, since Recursive Foreach is such a base-level method, there should be a flag passed to the method for whether or not to pay attention to the `_skip` set.
> 
> PLA-1628

Can you think of an application for skipping `_skip`?  I'm inclined to leave it out (I don't think it was ever intended), but I agree that it seems like a thing that could be useful in the future at some point.

kroenlein:
> Can you think of an application for skipping _skip? I'm inclined to leave it out (I don't think it was ever intended), but I agree that it seems like a thing that could be useful in the future at some point.

I have large interconnected set of objects that share an input material (salt).  I wish to perform some logic on all DAGs containing that input material.  It seemed like the intent of the `_skip` field was specifically to avoid serializing particular field, and any `_skip` behavior is a little non-intuitive for a method called `recursive_foreach` (as opposed to `recursive_forsome`)

taurus/util/tests/test_foreach.py
kroenlein: I can understand pushback on the final process, but if you do not at least have 1 full cycle, you are not testing all references.

maxhutch:
> I have large interconnected set of objects that share an input material (salt). I wish to perform some logic on all DAGs containing that input material. 

I don't follow.

> It seemed like the intent of the `_skip` field was specifically to avoid serializing particular field, and any `_skip` behavior is a little non-intuitive for a method called `recursive_foreach` (as opposed to `recursive_forsome`)

Yes, that's why I think we should not skip _skip by default.  The question is whether or not we should skip _skip optionally ever.

taurus/util/tests/test_foreach.py
bfolie: Isn't material -> ingredient -> process -> material a full cycle?

PR #58, opened 2019-12-12
Adding failing test for round-robin ingredients through json

maxhutch:
Add test coverage of ingredients being included in serialized processes.


maxhutch:
Looks like this was resolved with #78

PR #59, opened 2020-01-02
Don't mutate original object in substitute_objects

asantas93:
This change is a step in the direction of a totally immutable JSON deserialization system. Right now, mutating JSON is leading to the need for calls to `deepcopy` in this project and in `citrine-python`. Deep copies are expensive and introduce noticeable latency (about a second in a recent example) for large JSON payloads such as in a material histories.

### DIFF
--- a/taurus/util/impl.py
+++ b/taurus/util/impl.py
@@ -46,53 +46,39 @@ def substitute_objects(obj, index):
 
     This prepares the object to be used after being deserialized.
     It is the inverse of substitute_links.
-    It is an in-place operation.
     :param obj: target of the operation
     :param index: containing the objects that the uids point to
     :return: None
     """
-    visited = set()
+    visited = {}
+
+    def index_or_obj(x):
+        if isinstance(x, LinkByUID):
+            return index.get((x.scope.lower(), x.id), x)
+        else:
+            return x
 
     def substitute(thing):
         if id(thing) in visited:
-            return
-        visited.add(id(thing))
-        if isinstance(thing, (list, tuple)):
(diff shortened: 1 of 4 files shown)


taurus/util/tests/test_substitute_objects.py
```diff
@@ -19,8 +19,19 @@ def test_dictionary_substitution():
              (proc_link.scope.lower(), proc_link.id): proc}
 
     test_dict = {LinkByUID.from_entity(proc): LinkByUID.from_entity(mat)}
-    substitute_objects(test_dict, index)
-    assert test_dict[proc] == mat
```
asantas93: Since the resulting process run in `subbed` is not the literal exact same object anymore, this assertion as it was failed (we don't implement a reasonable hash function for our models).

taurus/util/impl.py
```diff
@@ -46,53 +46,39 @@ def substitute_objects(obj, index):
...
-                if isinstance(k, LinkByUID) and (k.scope.lower(), k.id) in index:
-                    thing[index[(k.scope.lower(), k.id)]] = v
-                    del thing[k]
-                else:
-                    substitute(k)
+            new = {substitute(index_or_obj(k)): substitute(index_or_obj(v))
+                   for k, v in thing.items()}
         elif isinstance(thing, DictSerializable):
-            for k, v in vars(thing).copy().items():
-                if isinstance(thing, BaseEntity) and k in thing.skip:
```
asantas93: `k in thing.skip` will always be `False` when iterating over the result of `as_dict`.

PR #60, opened 2020-01-03
Don't mutate object when subbing links

asantas93:
This is based on the changes in https://github.com/CitrineInformatics/taurus/pull/59. If we merge both, we ought to be able to simplify some serde logic in citrine-python (and almost certainly taurus as well).


taurus/client/json_encoder.py
```diff
@@ -153,7 +153,7 @@ def thin_dumps(obj, **kwargs):
     """
     set_uuids(obj)
     res = deepcopy(obj)
```
bfolie: Can we get rid of this deepcopy now that `substitute_links` does not mutate?

taurus/util/impl.py
```diff
@@ -25,19 +26,57 @@ def func(base_obj):
     return
 
 
+def _substitute(thing, sub, applies, visited: Dict[int, object] = None):
```
bfolie: Could you add a docstring for this method?
Also, why is the typehint for `visited` a Dict from int to object? Shouldn't it be `Dict[object, object]`?

asantas93: yeah that's from an old implementation, I will add a docstring and fix the type hint

taurus/client/json_encoder.py
asantas93: probably, I'll look around for `deepcopy`s to remove

PR #62, opened 2020-01-08
Cake update

kroenlein:
An update to add some untested elements for display.

* Custom UIDs
* UTF-8 / non-ASCII characters
* A FileLink on a Run

### DIFF
--- a/taurus/demo/cake.py
+++ b/taurus/demo/cake.py
@@ -191,7 +191,7 @@ def make_cake_spec(tmpl=None):
         name="Abstract Cake",
         template=tmpl["Dessert"],
         process=ProcessSpec(
-            name='Icing, in General',
+            name='Icing Cake, in General',
             template=tmpl["Icing"],
             tags=[
                 'spreading'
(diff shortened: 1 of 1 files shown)


taurus/demo/cake.py
```diff
@@ -754,14 +780,22 @@ def make_cake(seed=None, tmpl=None, cake_spec=None):
...
+    set_uuids([cake, cake_taste, cake_appearance,
+               frosting_taste, frosting_sweetness], name=DEMO_SCOPE)
     id_queue = [x for x in cake.process.ingredients]
     while id_queue:
         x = id_queue.pop(0)
-        set_uuids([x], name='id')
+        set_uuids([x], name=DEMO_SCOPE)
         id_queue += x.material.process.ingredients
 
+    cake.notes = cake.notes + "; Très délicieux! 😀"
```
bfolie: Whoa

kroenlein:
> I think that all of the id-applying code can be removed and replaced with a single call to `recursive_foreach` (a method in util/imply.py) using `cake` as the initial object and a function that takes an object and calls `obj.add_uid(DEMO_SCOPE, obj.name)`
> 
> _edit_: I tried `recursive_foreach(cake, lambda obj: obj.add_uid("another demo scope", obj.name))`, and it applied the custom id to every object.

Can you post the update here, or Slack me the code?  The previous version was largely legacy from when things didn't work as well.

taurus/demo/cake.py
```diff
@@ -154,6 +159,9 @@ def make_cake_templates():
     tmpl["Procurement"] = ProcessTemplate(name="Procurement",
                                           description="Buyin' stuff")
 
+    for key in tmpl:
+        tmpl[key].add_uid(DEMO_SCOPE, key)
+
```
bfolie: This can be removed. It's true that it is slightly different from the id assignment below, in that it uses the key in the template dictionary, which isn't necessarily equal to the name ("Generic Material" and "Baked Color", for example). But the ids assigned here are overwritten by the recursive_foreach, since they all use the same scope.

ventura-rivera:
@bfolie Ken and I talked about a hack to fix the UID problem temporarily since the FE team will be needing this soon. He is going to push those changes soon

PR #63, opened 2020-01-09
Reflect intent not to rename IngredientRuns

asantas93:
@ventura-rivera pointed out that this line was leading to misalignment of ingredient run and spec names in nextgen-devkit. This behavior was not observed in the dumped taurus cake example you get with `python taurus/demo/cake.py` because of [a call to loads](https://github.com/CitrineInformatics/taurus/blob/092feb89cad3d97bcf04549504cd75d7303cff18/taurus/entity/util.py#L115), which overwrites the `name` and `labels` again when `spec` is populated.

### DIFF
--- a/taurus/demo/cake.py
+++ b/taurus/demo/cake.py
@@ -637,7 +637,8 @@ def make_cake(seed=None, tmpl=None, cake_spec=None):
     queue = [cake]
     while queue:
         item = queue.pop(0)
-        item.name = item.name.replace('Abstract ', '').replace(', in General', '')
+        if not isinstance(item, IngredientRun):
+            item.name = item.name.replace('Abstract ', '').replace(', in General', '')
         if item.spec.tags is not None:
             item.tags = list(item.spec.tags)
         if item.spec.notes:  # None or empty string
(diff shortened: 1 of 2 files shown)


taurus/demo/cake.py
```diff
@@ -215,7 +215,7 @@ def make_cake_spec(tmpl=None):
         notes='Chocolate frosting'
     )
     IngredientSpec(
-        name="{} input".format(frosting.name),
+        name="{} input".format(frosting.name.replace('Abstract ', '')),
```
asantas93: Maybe make a utility function for these?

taurus/entity/object/ingredient_run.py
```diff
@@ -12,6 +12,25 @@ class IngredientRun(BaseObject, HasQuantities):
 
     Ingredients annotate a material with information about its usage in a process.
 
+    *  If a user invokes the IngredientRun constructor in citrine-python with either the name
+       or labels argument, a deprecation warning will be issued
```
bfolie: It feels wrong to reference citrine-python behavior in taurus. Similarly, bullets 3 & 4 refer to API calls and 400 codes, neither of which are relevant in this context.

taurus/demo/cake.py
kroenlein: Utility function for all 3 materials fields that were kwargs for the Ingredient constructors.

PR #64, opened 2020-01-13
Bug/PLA-2252 nominal composition as dictionary

andyczerwonka:
## [Composition representation doesn't follow spec](https://citrine.atlassian.net/browse/PLA-2252)

We were generating an array to represent nominal_composition pairs. This is not to spec. This PR removes that transformation, maintaining the dictionary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ coverage.xml
 *.cover
 .hypothesis/
 .pytest_cache/
+flex_measurements.json
 
 # Translations
 *.mo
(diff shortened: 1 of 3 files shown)


andyczerwonka:
I can't enable https://github.com/CitrineInformatics/nextgen-devkit/blob/3789fd6b06436a8e067dd4a3752daece76c99e0e/end-to-end-tests/tests/strike-team-data/test_template_validation.py#L156 until this is merged

PR #74, opened 2020-01-29
Migrate input to S&C example to citrination-derived PIF

kroenlein:
Migrated Strehlow and Cook Example to consume Citrination PIF as source
* Augmented dataset with omitted columns
* Verified inconsistencies against source document
* Generally made data interactions more abstract/general

To Do: 
* Include Computational v. Experimental data labels
* Include method annotations
* Import originally cited sources


taurus/demo/strehlow_and_cook.py
```diff
@@ -168,12 +286,11 @@ def make_strehlow_table(compounds):
...
     chem_mat_tmpl = compounds[0].spec.template
 
-    for comp in compounds:
-        if len(comp.measurements) == 3:  # Full list
-            band_msr = comp.measurements[0].spec
-            cryst_msr = comp.measurements[1].spec
-            color_msr = comp.measurements[2].spec
-            break
+    tmpl = dict()
+    # Horrible formatting to satisfy flake8
```
bfolie: Don't blame the linter. You could add lines
```
properties = compounds[0].measurements[0].spec.template.properties
conditions = compounds[0].measurements[0].spec.template.conditions
for attr in (properties + conditions):
```
Which is more readable, anyway

taurus/demo/strehlow_and_cook.py
```diff
@@ -232,94 +336,99 @@ def make_strehlow_table(compounds):
...
 
     return output
 
 
-if __name__ == "__main__":
-    import pkg_resources
-    import json
-    import taurus.client.json_encoder as je
+def make_display_table(structured):
+    """Generate a Display Table from a passed Structured Table."""
```
bfolie: What exactly is a structure table? Is there a "structured table" class? This would be more readable with some type hints, or an expanded docstring.

taurus/demo/tests/test_sac.py
```diff
@@ -1,23 +1,23 @@
...
-    for comp, row in zip(sac, sac_tbl['content']):
-        mask = ''.join(map(lambda x: str(type(x)), row))
-        if mask not in seen:
-            seen.add(mask)
-            # Check that all shapes of records serialize and deserialize
-            assert je.loads(je.dumps(comp)) == comp
+    smaller = minimal_subset(sac_tbl['content'])
     # Make sure that the diversity of value types isn't lost, e.g. something is being None'd
-    assert len(seen) == 12
+    assert len(smaller) == 153
```
bfolie: Is there a way to do this check without hardcoding 153? For example, can you count how many unique fingerprints are in the main table and use that number?

kroenlein: See response below

bfolie: I do not see a response below. Did you forget to submit the comment?

kroenlein: Not so much forget, as didn't get around to.  See comment below.

kroenlein:
> This is difficult to follow. I think it could benefit from a bit more documentation, and more importantly some enhanced tests. The tests right now ensure that everything can be loaded without error and that the resulting objects/table rows can be serialized and deserialized. There should be additional tests that verify the structure of the rows. I think that the test `assert len(smaller) == 153` is addressing this point, but it's unclear to me as a reader exactly what that test is ensuring. Perhaps you could verify that all templates appear in the objects, or have a hard-coded list of expected column headers and ensure that they all appear in the minimal subset.

The length test was guaranteeing that we weren't losing variety in a way that was lightweight and easily updated.  Watching the number of unique shapes drop is a good regression test for verifying the structure.  I expect in the not-so-distant future, moving this into a more strict regression test or removing table builds entirely makes sense as table construction is handled on the platform.  As it is, the small size of the SMALL_TABLE was motivated by a desire to keep test time down (the current test requires roughly 8 s for the full table, which is faster than it was but is too long for a unit test).

I agree that these changes need to get factored in, but I believe that the current improvements justify getting this PR through and moving additional improvements into another PR.

PR #75, opened 2020-02-03
Added measurement parameters to example

kroenlein:
Product/FE requested that parameters be added to measurement objects.

* This is still sensitive to the total-object-count validation, which serves as a regression test but may fail to be testing the right thing.
* Added the ParameterTemplate `Expected Sample Size`
* Added `Expected Sample Size` to the `Nutritional Analysis` and `Elemental Analysis` MeasurementTemplate
* Added Parameter `Expected Sample Size` to `Flour nutritional analysis`, `Salt elemental analysis`, `Sugar elemental analysis`, `Nutritional analysis` and `Elemental analysis` Measurement Runs and Measurement Specs

### DIFF
--- a/taurus/demo/cake.py
+++ b/taurus/demo/cake.py
@@ -118,15 +118,15 @@ def make_cake_templates():
             ]
         )
     )
-    tmpl["Sample Size"] = ConditionTemplate(
-        name="Sample Size",
+    tmpl["Sample Mass"] = ConditionTemplate(
+        name="Sample Mass",
         description="Sample size in mass units, to go along with FDA Nutrition Facts",
-        bounds=RealBounds(1.e-3, 10.e3, "g")
+        bounds=RealBounds(1.e-3, 1.e4, "g")
     )
-    tmpl["Expected Sample Size"] = ParameterTemplate(
-        name="Expected Sample Size",
+    tmpl["Expected Sample Mass"] = ParameterTemplate(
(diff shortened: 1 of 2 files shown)


taurus/demo/cake.py
```diff
@@ -123,6 +123,11 @@ def make_cake_templates():
         description="Sample size in mass units, to go along with FDA Nutrition Facts",
         bounds=RealBounds(1.e-3, 10.e3, "g")
     )
+    tmpl["Expected Sample Size"] = ParameterTemplate(
+        name="Expected Sample Size",
+        description="Specified sample size in mass units, to go along with FDA Nutrition Facts",
+        bounds=RealBounds(1.e-3, 10.e3, "g")
```
bfolie: Why 10.e3 instead of 1.e4?

taurus/demo/cake.py
```diff
@@ -123,6 +123,11 @@ def make_cake_templates():
         description="Sample size in mass units, to go along with FDA Nutrition Facts",
         bounds=RealBounds(1.e-3, 10.e3, "g")
     )
+    tmpl["Expected Sample Size"] = ParameterTemplate(
+        name="Expected Sample Size",
```
bfolie: "Expected Sample Size" sounds like it is the expected number of samples. Why not "Expected Sample Mass"?

PR #78, opened 2020-02-07
Include entire material histories when flattening, dumping, and loading

maxhutch:
None of those assertions seem to work, which is very weird.

### DIFF
--- a/taurus/entity/object/process_run.py
+++ b/taurus/entity/object/process_run.py
@@ -1,3 +1,5 @@
+from typing import Iterable
+
 from taurus.entity.object.base_object import BaseObject
 from taurus.entity.object.has_conditions import HasConditions
 from taurus.entity.object.has_parameters import HasParameters
@@ -107,3 +109,17 @@ class ProcessRun(BaseObject, HasConditions, HasParameters, HasSource):
             return self.spec.template
         else:
             return None
+
+    @property
+    def _forward(self) -> Iterable[str]:
+        """List of forward links that should be skipped when doing a chronological traversal.
+
+        Usually, this is the same as the skip list.  However, the link direction between
+        ingredients and processes is flipped: the ingredient -> process links is the writeable one
+        while the process -> ingredient link is read-only.  This is the opposite of chronological
+        order, so _forward gives a way to express the chronological ordering as well.
+
+        Ingredients are _not_ forward links, even though they are skipped, so they are not included
+        below.
+        """
+        return {"_output_material"}
--- a/taurus/entity/object/process_spec.py
+++ b/taurus/entity/object/process_spec.py
@@ -1,3 +1,5 @@
+from typing import Iterable
+
 from taurus.entity.object.base_object import BaseObject
 from taurus.entity.object.has_parameters import HasParameters
 from taurus.entity.object.has_conditions import HasConditions
@@ -82,3 +84,17 @@ class ProcessSpec(BaseObject, HasParameters, HasConditions, HasTemplate):
     def output_material(self):
         """Get the output material spec."""
         return self._output_material
+
+    @property
+    def _forward(self) -> Iterable[str]:
+        """List of forward links that should be skipped when doing a chronological traversal.
+
+        Usually, this is the same as the skip list.  However, the link direction between
+        ingredients and processes is flipped: the ingredient -> process links is the writeable one
+        while the process -> ingredient link is read-only.  This is the opposite of chronological
+        order, so _forward gives a way to express the chronological ordering as well.
+
+        Ingredients are _not_ forward links, even though they are skipped, so they are not included
+        below.
+        """
+        return {"_output_material"}
--- a/taurus/tests/test_examples.py
+++ b/taurus/tests/test_examples.py
@@ -51,7 +51,7 @@ def make_data_island(density, bulk_modulus, firing_temperature, binders, powders
     binder_runs = keymap(lambda x: MaterialRun(spec=x), binder_specs)
     powder_runs = keymap(lambda x: MaterialRun(spec=x), powder_specs)
 
-    all_input_materials = merge(binder_runs, powder_runs)
+    all_input_materials = keymap(lambda x: x.spec.name, merge(binder_runs, powder_runs))
     mixing_composition = Condition(
         name="composition",
         value=NominalComposition(all_input_materials)
@@ -157,9 +157,9 @@ def test_access_data():
     # read the quantity of alumina
     quantities = island.process.ingredients[0].material.process.conditions[0].value.quantities
     assert(list(
-        keyfilter(lambda x: x.spec.name == "Al2O3", quantities).values()
+        keyfilter(lambda x: x == "Al2O3", quantities).values()
     )[0] == 0.96)
 
     # check that the serialization results in the correct number of objects in the preface
     # (note that neither measurements nor ingredients are serialized)
-    assert(len(json.loads(dumps(island))[0]) == 8)
+    assert(len(json.loads(dumps(island))[0]) == 23)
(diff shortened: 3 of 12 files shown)


taurus/util/impl.py
```diff
@@ -181,7 +189,7 @@ def recursive_foreach(obj, func, apply_first=False, seen=None):
     return
 
 
-def recursive_flatmap(obj, func, seen=None):
+def recursive_flatmap(obj, func, seen=None, chronological=False):
```
gregor-robinson: Suggestion template:
`@param chronological causes func to be called in reverse chronological order, rather than the default _____ ordering`

taurus/tests/test_examples.py
```diff
@@ -51,7 +51,7 @@ def make_data_island(density, bulk_modulus, firing_temperature, binders, powders
     binder_runs = keymap(lambda x: MaterialRun(spec=x), binder_specs)
     powder_runs = keymap(lambda x: MaterialRun(spec=x), powder_specs)
 
-    all_input_materials = merge(binder_runs, powder_runs)
+    all_input_materials = keymap(lambda x: x.spec.name, merge(binder_runs, powder_runs))
```
maxhutch: Composition keys have to be strings, not material specs.

taurus/util/impl.py
```diff
@@ -57,7 +62,7 @@ def _substitute(thing,
     elif isinstance(thing, DictSerializable):
         new_attrs = {_substitute(k, sub, applies, visited): _substitute(v, sub, applies, visited)
                      for k, v in thing.as_dict().items()}
-        new = thing.from_dict(new_attrs)
+        new = thing.build(new_attrs)
```
maxhutch: This is what fixes https://github.com/CitrineInformatics/citrine-python/pull/203 as far as I can tell

taurus/util/impl.py
```diff
@@ -136,8 +150,8 @@ def _flatten(base_obj):
 
         return to_return
 
-    res = recursive_flatmap(obj, _flatten)
-    return [substitute_links(x) for x in res]
+    res = recursive_flatmap(obj, _flatten, chronological=True)
+    return sorted([substitute_links(x) for x in res], key=lambda x: _flattened_sort_order(x))
```
maxhutch: This sorting is how we can re-link the objects when being loaded from the preamble.

taurus/util/impl.py
```diff
@@ -203,24 +219,46 @@ def recursive_flatmap(obj, func, seen=None):
...
-                res.extend(recursive_flatmap(x, func, seen))
+                res.extend(recursive_flatmap(x, func, seen, chronological))
                 res.extend(func(x))
             else:
-                res.extend(recursive_flatmap(x, func, seen))
+                res.extend(recursive_flatmap(x, func, seen, chronological))
     return res
+
+
+def _flattened_sort_order(obj: BaseEntity) -> int:
```
maxhutch: This should be consistent with the backend write ordering https://github.com/CitrineInformatics/platform-backend/blob/0b958e2812d6b4e5a7c37ff3d1e2f89776a42c88/code/faraday/model/src/main/scala/io/citrine/faraday/api/model/Storable.scala#L264

maxhutch:
Confirmed that this fixes https://github.com/CitrineInformatics/citrine-python/pull/203 : 
https://travis-ci.org/CitrineInformatics/citrine-python/jobs/648027943

taurus/util/tests/test_flatten.py
```diff
@@ -0,0 +1,54 @@
...
+    procured = ProcessSpec(name="procured")
+    input = MaterialSpec(name="foo", process=procured)
+    transform = ProcessSpec(name="transformed")
+    ingredient = IngredientSpec(name="input", material=input, process=transform)
+
+    procured_run = ProcessRun(name="procured", spec=procured)
+    input_run = MaterialRun(name="foo", process=procured_run, spec=input)
+    transform_run = ProcessRun(name="transformed", spec=transform)
+    ingredient_run = IngredientRun(
+        name="input", material=input_run, process=transform_run, spec=ingredient)
```
kroenlein:
```suggestion
```

taurus/util/tests/test_flatten.py
```diff
@@ -0,0 +1,54 @@
...
+    """Test that flatten works when the objects are empty and go through a whole history."""
+    procured = ProcessSpec(name="procured")
+    input = MaterialSpec(name="foo", process=procured)
+    transform = ProcessSpec(name="transformed")
+    ingredient = IngredientSpec(name="input", material=input, process=transform)
+
+    procured_run = ProcessRun(name="procured", spec=procured)
+    input_run = MaterialRun(name="foo", process=procured_run, spec=input)
+    transform_run = ProcessRun(name="transformed", spec=transform)
+    ingredient_run = IngredientRun(
```
kroenlein: `name` parameter is vestigial for IngredientRuns

```suggestion
    ingredient_run = IngredientRun(material=input_run, process=transform_run, spec=ingredient)
```

taurus/util/tests/test_flatten.py
```diff
@@ -0,0 +1,54 @@
...
+    assert len(flatten(ingredient_run)) == 7
+    assert len(flatten(transform_run)) == 7
+
+
+def test_flatmap_skip_ordering():
+    """Test that the chronological setting is obeyed."""
+    # The writeable link is ingredient -> process, but the chronological link is
+    # process -> ingredient
+    proc = ProcessRun(name="foo")
+    IngredientRun(name="bar", process=proc)
```
kroenlein: Keep the foo/bar notation, but keep the constructor from issuing spurious warnings.
```suggestion
    IngredientRun(notes="bar", process=proc)
```

taurus/util/tests/test_flatten.py
kroenlein: (Multi-line suggestion)

taurus/entity/object/process_run.py
```diff
@@ -107,3 +109,17 @@ def template(self):
             return self.spec.template
         else:
             return None
+
+    @property
+    def _forward(self) -> Iterable[str]:
+        """List of forward links that should be skipped when doing a chronological traversal.
+
+        Usually, this is the same as the skip list.  However, the link direction between
```
gregor-robinson: Referencing the documentation of `BaseEntity._forward` here, rather than duplicating it, might reduce the future cost of avoiding diverging documentation.

taurus/entity/object/process_spec.py
```diff
@@ -82,3 +84,17 @@ def _unset_ingredient(self, ingred):
     def output_material(self):
         """Get the output material spec."""
         return self._output_material
+
+    @property
+    def _forward(self) -> Iterable[str]:
+        """List of forward links that should be skipped when doing a chronological traversal.
+
+        Usually, this is the same as the skip list.  However, the link direction between
```
gregor-robinson: Referencing the documentation of `BaseEntity._forward` here, rather than duplicating it, might reduce the future cost of avoiding diverging documentation.

taurus/util/impl.py
gregor-robinson: Is this documented in the API spec, at least as a hint to keep the ordering synchronized between backend and client?

taurus/util/impl.py
```diff
@@ -112,6 +118,11 @@ def flatten(obj):
       - Making sure at least one uid is set in each BaseEntity in scope
       - Getting a list of unique objects contained in the scope
       - Substituting the pointers in those objects with LinkByUID objects
```
gregor-robinson:
```suggestion
      - Substituting the pointers in those objects with LinkByUID objects
      - Sorting the output so an object is listed after all of its dependencies
```

taurus/entity/object/process_spec.py
maxhutch: Given that we're currently spending most of our time in-code rather than in sphinx docs, I think this duplication is warranted.

taurus/util/impl.py
maxhutch: No; it is an implicit ordering given other documentation about the way bidirectional links in the data model documentation are bound in the API.

PR #84, opened 2020-02-17
Bugfix: json_encoder.dump was dropping material -> measurement links

maxhutch:
When we dump a MaterialRun, we want to dump any of its connected
MeasurementRuns as well (at least nominally).

### DIFF
--- a/taurus/entity/object/material_run.py
+++ b/taurus/entity/object/material_run.py
@@ -126,4 +126,11 @@ class MaterialRun(BaseObject):
 
     @property
     def _forward(self) -> Iterable[str]:
+        """List of forward links that should be skipped when doing a chronological traversal.
+
+        There are no forward links present in the material run.  The only item in skip,
+        measurements, should be included when doing a "chronological" traversal.  It just
+        shouldn't be included *as a member* when serializing the material run objects.
+        :return: an empty set
+        """
         return {}
--- a/taurus/entity/object/material_run.py
+++ b/taurus/entity/object/material_run.py
@@ -1,3 +1,5 @@
+from typing import Iterable
+
 from taurus.entity.object.base_object import BaseObject
 from taurus.enumeration import SampleType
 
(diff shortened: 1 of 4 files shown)


maxhutch:
@kroenlein @bfolie : this represents a real change in expectation where `json_encoder.dump` will now include measurements on materials.

taurus/tests/test_examples.py
```diff
@@ -162,4 +162,4 @@ def test_access_data():
 
     # check that the serialization results in the correct number of objects in the preface
     # (note that neither measurements nor ingredients are serialized)
-    assert(len(json.loads(dumps(island))[0]) == 23)
+    assert(len(json.loads(dumps(island))[0]) == 26)
```
kroenlein: Brendan has rightly pointed out that element count tests are weak.

taurus/entity/object/material_run.py
```diff
@@ -121,3 +123,7 @@ def template(self):
             return self.spec.template
         else:
             return None
+
+    @property
+    def _forward(self) -> Iterable[str]:
+        return {}
```
kroenlein:
```suggestion
        """List of forward links that should be skipped when doing a chronological traversal.

        Usually, this is the same as the skip list.  However, the link direction between
        ingredients and processes is flipped: the ingredient -> process links is the writeable one
        while the process -> ingredient link is read-only.  This is the opposite of chronological
        order, so _forward gives a way to express the chronological ordering as well.

        Ingredients are _not_ forward links, even though they are skipped, so they are not included
        below.
        """
        return {}
```

maxhutch: I'll update that

taurus/tests/test_examples.py
maxhutch: True, but I think that's out of scope.

maxhutch:
Note that this change doesn't change the result of `as_dict()`: we still doesn't serialize links to the measurements in the material run.  What this changes is what is included in the context part of the json serialization, such that any connected measurements are ultimately deserialized, built, and at that time linked to the materials that they reference.

PR #86, opened 2020-02-18
Encapsulate json config in a class; support extensibility

maxhutch:
This way, the class list can be manipulated safely, since there's
no global state.  A default instance is created in the taurus.json
module that's used to provide taurus.json.load and
taurus.json.dump

### DIFF
--- a/taurus/client/json_encoder.py
+++ b/taurus/client/json_encoder.py
@@ -1,275 +1,53 @@
+# flake8: noqa
 """Encoding and decoding taurus objects to/from json."""
-import inspect
-import json
-
-from taurus.entity.attribute.condition import Condition
-from taurus.entity.attribute.parameter import Parameter
-from taurus.entity.attribute.property import Property
-from taurus.entity.attribute.property_and_conditions import PropertyAndConditions
-from taurus.entity.base_entity import BaseEntity
-from taurus.entity.bounds.categorical_bounds import CategoricalBounds
-from taurus.entity.bounds.composition_bounds import CompositionBounds
-from taurus.entity.bounds.integer_bounds import IntegerBounds
-from taurus.entity.bounds.real_bounds import RealBounds
-from taurus.entity.dict_serializable import DictSerializable
-from taurus.entity.file_link import FileLink
-from taurus.entity.link_by_uid import LinkByUID
-from taurus.entity.object import ProcessRun, MaterialRun, MeasurementRun
-from taurus.entity.object.ingredient_run import IngredientRun
-from taurus.entity.object.ingredient_spec import IngredientSpec
-from taurus.entity.object.material_spec import MaterialSpec
-from taurus.entity.object.measurement_spec import MeasurementSpec
-from taurus.entity.object.process_spec import ProcessSpec
-from taurus.entity.source.performed_source import PerformedSource
-from taurus.entity.template.condition_template import ConditionTemplate
-from taurus.entity.template.material_template import MaterialTemplate
-from taurus.entity.template.measurement_template import MeasurementTemplate
-from taurus.entity.template.parameter_template import ParameterTemplate
-from taurus.entity.template.process_template import ProcessTemplate
-from taurus.entity.template.property_template import PropertyTemplate
-from taurus.entity.value.discrete_categorical import DiscreteCategorical
-from taurus.entity.value.empirical_formula import EmpiricalFormula
-from taurus.entity.value.nominal_categorical import NominalCategorical
-from taurus.entity.value.nominal_composition import NominalComposition
-from taurus.entity.value.nominal_integer import NominalInteger
-from taurus.entity.value.nominal_real import NominalReal
-from taurus.entity.value.normal_real import NormalReal
-from taurus.entity.value.uniform_integer import UniformInteger
-from taurus.entity.value.uniform_real import UniformReal
-from taurus.enumeration.base_enumeration import BaseEnumeration
-from taurus.json import TaurusEncoder
-from taurus.util import flatten, substitute_links, set_uuids
+import deprecation
+import taurus
+from taurus.json import TaurusEncoder, TaurusJson
 
 
+@deprecation.deprecated(deprecated_in="0.6.0", removed_in="0.7.0",
+                        details="Use taurus.json.dumps instead")
 def dumps(obj, **kwargs):
-    """
-    Serialize a taurus object, or container of them, into a json-formatting string.
-
-    Parameters
-    ----------
-    obj: DictSerializable or List[DictSerializable]
-        The object(s) to serialize to a string.
-    **kwargs: keyword args, optional
-        Optional keyword arguments to pass to `json.dumps()`.
-
-    Returns
-    -------
-    str
-        A string version of the serialized objects.
-
-    """
-    # create a top level list of [flattened_objects, link-i-fied return value]
-    res = [obj]
-    additional = flatten(res)
-    res = substitute_links(res)
-    res.insert(0, additional)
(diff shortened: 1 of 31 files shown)


maxhutch:
Check out https://github.com/CitrineInformatics/taurus/blob/feature/encapsulate-json-config-in-class/docs/source/depth/serialization.rst for in-depth documentation.

tox.ini
```diff
@@ -17,5 +17,6 @@ ignore = D100,D104,D105,D107,D301,D401,E221
 omit =
     # omit demos
     taurus/demo/*
+    taurus/client/json_encoder.py
```
maxhutch: This file is deprecated, so we omit a bunch of docstrings.

taurus/json/taurus_json.py
```diff
@@ -0,0 +1,322 @@
...
+            Deserialized versions of the objects represented by `json_str`
+
+        """
+        # Create an index to hold the objects by their uid reference
+        # so we can replace links with pointers
+        index = {}
+        return json_builtin.loads(
+            json_str, object_hook=lambda x: self._load_and_index(x, index), **kwargs)
+
+    def register_classes(self, classes):
```
maxhutch: This is what citrine-python uses to inject its versions of the data objects and templates https://github.com/CitrineInformatics/citrine-python/pull/223/files#diff-67c951dbc53134cb68be92fa81d2d77dR161

taurus/client/json_encoder.py
```diff
@@ -1,269 +1,53 @@
...
-from taurus.entity.value.nominal_categorical import NominalCategorical
-from taurus.entity.value.nominal_composition import NominalComposition
-from taurus.entity.value.nominal_integer import NominalInteger
-from taurus.entity.value.nominal_real import NominalReal
-from taurus.entity.value.normal_real import NormalReal
-from taurus.entity.value.uniform_integer import UniformInteger
-from taurus.entity.value.uniform_real import UniformReal
-from taurus.enumeration.base_enumeration import BaseEnumeration
-from taurus.util import flatten, substitute_links, set_uuids
+import deprecation
```
kroenlein: Is the inclusion of `deprecation` conflating a separate issue into this proposal?

taurus/demo/cake.py
```diff
@@ -925,41 +925,43 @@ def find_name(name, material):
...
+    encoder = TaurusJson()
     cake = make_cake(seed=42)
 
     with open("example_taurus_material_history.json", "w") as f:
         context_list = complete_material_history(cake)
         f.write(json.dumps(context_list, indent=2))
 
     with open("example_taurus_material_template.json", "w") as f:
-        f.write(thin_dumps(cake.template, indent=2))
+        f.write(encoder.thin_dumps(cake.template, indent=2))
```
kroenlein: Are we still using these example dumps elsewhere?  I thought they were feeding a platform test or development target.

taurus/entity/object/tests/test_material_run.py
```diff
@@ -34,7 +34,7 @@ def test_material_run():
 
     # Make sure that when property is serialized, origin (an enumeration) is serialized as a string
     copy_prop = json.loads(dumps(mat_spec))
-    copy_origin = copy_prop[0][0]["properties"][0]['property']['origin']
+    copy_origin = copy_prop["context"][0]["properties"][0]['property']['origin']
```
kroenlein: This is a much better syntax that results from the shift.

taurus/json/taurus_json.py
```diff
@@ -0,0 +1,322 @@
...
+from taurus.entity.template.property_template import PropertyTemplate
+from taurus.entity.value.discrete_categorical import DiscreteCategorical
+from taurus.entity.value.empirical_formula import EmpiricalFormula
+from taurus.entity.value.nominal_categorical import NominalCategorical
+from taurus.entity.value.nominal_composition import NominalComposition
+from taurus.entity.value.nominal_integer import NominalInteger
+from taurus.entity.value.nominal_real import NominalReal
+from taurus.entity.value.normal_real import NormalReal
+from taurus.entity.value.uniform_integer import UniformInteger
+from taurus.entity.value.uniform_real import UniformReal
```
kroenlein: For the scale of this list and the nature of the import, wouldn't it make more sense to `import *`?  If those imports auto-registered the modules with _clazzes it then we remove fragility.

maxhutch: > For the scale of this list and the nature of the import, wouldn't it make more sense to `import *`?

Sure, but we don't have the imports to support that: https://github.com/CitrineInformatics/taurus/blob/develop/taurus/entity/value/__init__.py

I'd like to treat that as a separate issue.

> If those imports auto-registered the modules with _clazzes it then we remove fragility.

I'm not sure what you mean here.  Are you suggesting reflecting into the members of a module?

taurus/demo/cake.py
maxhutch: Yes, when we make data model changes we use this to update some tests in platform-backend.

taurus/client/json_encoder.py
maxhutch: We moved these methods into `taurus.json` so this just deprecates their old location.  `taurus.client.json_encoder` was a vestige from much earlier development (when I had envisioned implementing a whole local client).

taurus/json/__init__.py
```diff
@@ -0,0 +1,86 @@
+from .taurus_encoder import TaurusEncoder  # noqa: F401
+from .taurus_json import TaurusJson
+
+__default = TaurusJson()
```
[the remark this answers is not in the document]
maxhutch: It is meant as a drop-in replacement for the `json` module provided by python, mirroring the methods that it provides.  I think most users will use this.  I would expect developers (like us when working in citrine-python) to create `TaurusJson` instances.  I'll add a docstring reflecting this.

taurus/json/taurus_json.py
```diff
@@ -0,0 +1,322 @@
...
+    material run and a measurement run by setting the `material` field in the measurement run,
+    but not the other way around; the measurements field in material run is read-only.  This
+    is reflected in the json strategy through the `skip` member of DictSerializable, which
+    makes sure that the read-only side of bidirectional links are not written.
+
+    The second part of the solution to (a) is a seen set in the `recursive_flatmap` method
+    in util that is used by `flatten`.
+
+    The (b) problem is resolved by those same `recursive_flatmap` and `flatten` methods.  The
+    `flatten` method does a depth-first traversal of the graph, recording the first time
```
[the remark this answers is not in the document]
maxhutch: Just once.  I'll clarify.

taurus/json/__init__.py
```diff
@@ -19,7 +37,7 @@ def loads(json_str, **kwargs):
     -------
     DictSerializable or List[DictSerializable]
         Deserialized versions of the objects represented by `json_str`, with links turned
-        back into pointers.
+        back into object references (python's pointers).
```
gregor-robinson:
```suggestion
        back into python object references.
```

gregor-robinson: Python implements references with pointers under the hood, but the word pointer suggests a need for explicit dereference, which is the connotation I was trying to avoid.

taurus/json/taurus_json.py
kroenlein: Not entire sure if it's textbook reflection, but basically yes.  

There are a few ways to do it, but it would mostly make sense if taurus/json were tied strongly to DictSerializable, as opposed to being an external module that crawls its trees.

maxhutch: Could you give an example of a way to do it?  If this is merged by then, then you could do it in another PR.  Seems like a relatively localized change.

PR #87, opened 2020-02-19
Include the output material when serializing processes

maxhutch:
This acknolwedges that (process, output_material) are two halves
of the same chronological entity.  As a positive side-effect,
this means that every object that is accessible via the
python attributes of the objects is included in serialization,
which removes the need for the _forward option and simplifies the code.

To reflect the singular purpose of skip, the option in recursive_flatmap
has been renamed to "unidirectional": only the writable half of
bidirectional links are traversed when unidirection is True.

### DIFF
--- a/taurus/entity/base_entity.py
+++ b/taurus/entity/base_entity.py
@@ -1,6 +1,4 @@
 """Base class for all entities."""
-from typing import Iterable
-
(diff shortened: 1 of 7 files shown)


maxhutch:
Note: this works locally with https://github.com/CitrineInformatics/citrine-python/pull/223

PR #92, opened 2020-03-07
Adpated S&C demo to named instead of indexed json structure

kroenlein:
The shift from array to dictionary representation broke some code in the name==main block of the Strehlow and Cook example.  This corrects that issue and adds a test to catch if a migration will cause a similar problem.  

It's still a mock that needs to be verified against actual structured data output, probably as a regression test.

### DIFF
--- a/taurus/demo/tests/test_sac.py
+++ b/taurus/demo/tests/test_sac.py
@@ -2,6 +2,7 @@
 from taurus.demo.strehlow_and_cook import make_strehlow_table, make_strehlow_objects, \
     minimal_subset, import_table
 import taurus.json as je
+import json
 
 
 def test_sac():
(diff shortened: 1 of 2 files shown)


taurus/demo/tests/test_sac.py
```diff
@@ -21,3 +22,6 @@ def test_sac():
     # Make sure there's no migration with repeated serialization
     for row in sac_tbl:
         assert je.dumps(je.loads(je.dumps(row))) == je.dumps(row)
+
+    # Verify that the serialization trick for mocking a structured table works
+    json.dumps(json.loads(je.dumps(sac_tbl))["object"], indent=2)
```
bfolie: What does it mean to verify that it works? Is it sufficient that this this object can be dumped without error or is there something else being tested?

kroenlein: `verify that it works` literally means doesn't throw an exception.  Not great form, but we're about to obsolete this pathway to tables.

PR #94, opened 2020-04-01
Setting the repo name to GEMD-python; deploy docs on tags

maxhutch:


docs/source/depth/unit_parsing.rst
```diff
@@ -4,11 +4,11 @@ Unit Parsing
...
 Unit parsing is performed using the Pint_ package.
 By default, Pint supports a larger set of units than the Citrine Platform.
-Therefore, we include a custom unit definition file in GEMD: `citrine_en.txt`_.
+Therefore, we include a custom unit definition file in GEMD-python: `citrine_en.txt`_.
 This file contains the most commonly used units and will grow over time.
 
-Requests for support of additional units can be made by opening an issue in the `GEMD repository`_ on github.
+Requests for support of additional units can be made by opening an issue in the `GEMD-python repository`_ on github.
 
 .. _Pint: https://pint.readthedocs.io/en/0.9/
```
kroenlein, line 12: Side note: `pint` 0.11.0 changed how it handles delta-T units in code, and so our units file is incompatible.  We should probably boost our required version, or possibly do some more careful version definition.

maxhutch: Can you write an issue for this?

PR #100, opened 2020-05-01
Add molecular structure support

kroenlein:
Adding support to GEMD for SMILES- and InChI-formatted molecular structures.

As per the designs communicated in 

https://citrineinformatics.github.io/gemd-docs/specification/value-types/

this implements and tests those structures.

https://citrine.atlassian.net/browse/PLA-1876


gemd/entity/value/smiles_value.py
```diff
@@ -0,0 +1,35 @@
+"""An empirical chemical formaula."""
+from gemd.entity.value.molecular_value import MolecularValue
+
+
+class SMILES(MolecularValue):
```
kroenlein, line 5: @djack What is Citrine Best Practice for acronyms in class names?  Should this be `SMILES`, `Smiles` or other?

sparadiso: It's `Smiles` in the backend, fwiw.

setup.py
```diff
@@ -36,7 +36,7 @@ def run(self):
 
 
 setup(name='gemd',
-      version='0.6.6',
+      version='0.6.7',
```
bfolie, line 39: Minor version bump because it's a new feature?

kroenlein: The deprecation warnings on the Taurus migration are:
```
./gemd/client/json_encoder.py:@deprecation.deprecated(deprecated_in="0.6.0", removed_in="0.7.0",
```
So, from a process perspective, I should be wiping out the Taurus tree in a different PR if we're doing the minor version bump.

bfolie: ack

Well I don't know what the best process is... but this PR itself looks fine.

PR #101, opened 2020-05-01
Added consistency between client and server implementations

kroenlein:
typ strings differed between client and server implementations.  This reconciles that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ class PostDevelopCommand(develop):
 
 
 setup(name='gemd',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -36,7 +36,7 @@ def run(self):
 
 
 setup(name='gemd',
-      version='0.7.0',
+      version='0.7.1',
```
maxhutch, line 39: Since you changed a class name, I think this should have been a minor version bump.  Not worth back-porting; just an fyi.

PR #102, opened 2020-05-04
Add molecular structure types to cake example

kroenlein:
Molecular structure types were added to the cake example.  In particular, for the two items that already had expected molecular structures, a SMILES was added to sugar and an InChI was added to salt, both using the same Molecular Structure bounds object.

This is expected to be a proper release.

https://citrine.atlassian.net/browse/PLA-3934

### DIFF
--- a/gemd/demo/cake.py
+++ b/gemd/demo/cake.py
@@ -685,9 +685,10 @@ def make_cake(seed=None, tmpl=None, cake_spec=None):
                                std=0.05 * item.spec.absolute_quantity.nominal,
                                units=item.spec.absolute_quantity.units)
             if item.spec.volume_fraction is not None:
+                # The only element here is dry mix, and it's almost entirely flour
                 item.volume_fraction = \
-                    NormalReal(mean=fuzz * item.spec.volume_fraction.nominal,
-                               std=0.05 * item.spec.volume_fraction.nominal,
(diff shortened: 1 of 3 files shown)


gemd/demo/cake.py
```diff
@@ -113,6 +115,11 @@ def make_cake_templates():
         description="The chemical formula of a material",
         bounds=CompositionBounds(components=EmpiricalFormula.all_elements())
     )
+    tmpl["Molecular Structure"] = PropertyTemplate(
+        name="Molecular Structure",
+        description="The molecular structure of the material",
+        bounds=MolecularStructureBounds()
+    )
```
bfolie, line 122: How is this template used?

kroenlein: It appears (or should be appearing) as a PropertyAndCondition on the MaterialSpec for salt and for sugar.  The former is represented as an InChI, the latter as a SMILES.

bfolie: How are the `PropertyAndCondition` objects linked back to the template? I see, for example,
`Property(name='SMILES', value=Smiles("C(C1C(C(C(C(O1)OC2(C(C(C(O2)CO)O)O)CO)O)O)O)O")))`, but shouldn'T that property be linked to the `PropertyTemplate`?

kroenlein: Yes.  Yes it should.

PR #104, opened 2020-05-08
Feature/walk gemd tree

ventura-rivera:
### Summary
* This functionality allows you to traverse through a terminal MaterialRun and pick out all the objects and templates into a set.
* This is particularly helpful on the DE side because we will need all of those objects and templates to perform a bulk write to the platform (or we want access to all of the objects for some other reason).
* A slightly different version (with the same results) is used in nextgen-devkit
* I find myself copying and pasting this around a lot (and @sesevgen has needed to use it too)

### DIFF
--- a/gemd/util/impl.py
+++ b/gemd/util/impl.py
@@ -7,6 +7,8 @@ from gemd.entity.dict_serializable import DictSerializable
 from gemd.entity.link_by_uid import LinkByUID
 from toolz import concatv
 
+from gemd.entity.object.material_run import MaterialRun
+
 
 def set_uuids(obj, name="auto"):
     """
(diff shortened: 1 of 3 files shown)


gemd/util/impl.py
```diff
@@ -265,3 +267,106 @@ def writable_sort_order(key: Union[BaseEntity, str]) -> int:
...
+def walk_gemd_tree(material: MaterialRun):
+    """Walks through a root material node to build a list of GEMD objects."""
+    from gemd.entity.object import MaterialRun, ProcessRun, MeasurementRun, IngredientRun
+    from gemd.entity.attribute.property_and_conditions import PropertyAndConditions
+    from gemd.entity.template.condition_template import ConditionTemplate
+    from gemd.entity.template.parameter_template import ParameterTemplate
+    from gemd.entity.template.property_template import PropertyTemplate
+    from gemd.entity.attribute.condition import Condition
+    from gemd.entity.attribute.parameter import Parameter
+    from gemd.entity.attribute.property import Property
```
ventura-rivera, line 281: Here's an example

maxhutch:
@ventura-rivera could you use [`flatten`](https://github.com/CitrineInformatics/gemd-python/blob/master/gemd/util/impl.py#L108) instead?  I think the difference is just that flatten also puts in `LinkByUID` to replace references to other objects, but that has to happen anyways to write the objects (citrine-python just does it automatically).

ventura-rivera:
> @ventura-rivera could you use [`flatten`](https://github.com/CitrineInformatics/gemd-python/blob/master/gemd/util/impl.py#L108) instead? I think the difference is just that flatten also puts in `LinkByUID` to replace references to other objects, but that has to happen anyways to write the objects (citrine-python just does it automatically).

Hey @maxhutch I just tried `flatten` out by replacing it in one of my projects that requires walking through 249 cakes, but it is much slower. It takes ~20 seconds vs the ~3 seconds it takes me with this function. Also, it is not returning the expected number of objects (I am expecting 32,370 objects but it is returning 32,121 objects).

kroenlein:
WIth #142, the performance concerns should be resolved.  Bulk load has been implemented in citrine-python.  Closing as resolved.  @ventura-rivera, if the explicit SDK interfaces are still of high value, we should open a ticket to add those elements.

PR #105, opened 2020-05-27
Remove unnecessary enumeration library (>=3.5)

kroenlein:
With the migration to supporting 3.5 and later, we no longer need the enum34 library.  The library was still listed in the requirements files, so this removes those obsolete lines with no change in functionality.


kroenlein:
It's not clear to me whether this will require a change to citrine-python, where the Pipfile.lock includes a reference to enum34.

PR #106, opened 2020-05-28
Added control for default scope in serialization

kroenlein:
This commit adds a parameter to the GEMDJson constructor for what the default scope for serialized objects should be.  This breaks the pattern of the Json constructor being argument free and formatting arguments being passed directly to the dumps, etc. methods, however it made more sense from a scoping perspective to have this be associated with the constructor rather than each individual invocation.


gemd/util/impl.py
```diff
@@ -105,7 +105,7 @@ def substitute_objects(obj, index):
                        applies=lambda o: isinstance(o, LinkByUID))
 
 
-def flatten(obj):
+def flatten(obj, scope):
```
kroenlein, line 108: This argument order can be swapped if we want; I think we were discussing forcing named arguments in general instead.

gemd/json/gemd_json.py
```diff
@@ -65,7 +67,8 @@ class GEMDJson(object):
 
     _link_type = LinkByUID
 
-    def __init__(self):
+    def __init__(self, scope='auto'):
+        self._scope = scope
```
maxhutch, line 71: Using the `self._var` pattern is usually accompanied with a `@property` declaration somewhere so `GemdJson(scope="foo").scope` is a thing that can be done. If you don't mind not controlling the setter/getter, you can just do `self.scope = scope`.

kroenlein: v. reasonable.  I wasn't sure we wanted it user visible at all, but that's pretty paternalistic.

PR #107, opened 2020-06-04
Modified unit defaults to reduce risks

kroenlein:
This reduces the number of short aliases for units in an attempt to be a bit less likely to yield unfortunate results.  Motivated by an attempt to enter centipoise (cP) and getting cups.  

When a line includes an underscore, the first term in the line is used to describe the unit.  For example:

`unified_atomic_mass_unit = atomic_mass_constant = _ = amu`

will represent a value as `5.0 unified_atomic_mass_unit` and display that upon request.  When the underscore is not present, the third term is used as the alias: 

`dalton = atomic_mass_constant = Da`

will represent a value as `5.0 Da`.

### DIFF
--- a/gemd/units/citrine_en.txt
+++ b/gemd/units/citrine_en.txt
@@ -339,7 +339,7 @@ bar = 1e5 * pascal
 technical_atmosphere = kilogram * g_0 / centimeter ** 2 = at
 torr = atm / 760
 pound_force_per_square_inch = force_pound / inch ** 2 = psi
-kip_per_square_inch = kip / inch ** 2 = _ = kip_per_square_inch
+kip_per_square_inch = kip / inch ** 2 = ksi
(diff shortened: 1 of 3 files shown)


gemd/units/citrine_en.txt
```diff
@@ -339,7 +339,7 @@ bar = 1e5 * pascal
 technical_atmosphere = kilogram * g_0 / centimeter ** 2 = at
 torr = atm / 760
 pound_force_per_square_inch = force_pound / inch ** 2 = psi
-kip_per_square_inch = kip / inch ** 2 = ksi
+kip_per_square_inch = kip / inch ** 2 = _ = kip_per_square_inch
```
kroenlein, line 342:
```suggestion
kip_per_square_inch = kip / inch ** 2 = ksi
```
Regress this change following DE feedback.

PR #108, opened 2020-06-11
PLA-2151 make bounds optional on object templates

asantas93:
* Bounds are now optional on object templates
* Bounds are set to `None` when unspecified instead of to that of the attribute template

### DIFF
--- a/gemd/entity/template/tests/test_process_template.py
+++ b/gemd/entity/template/tests/test_process_template.py
@@ -45,7 +45,7 @@ def test_allowed_labels():
 
 
 def test_passthrough_bounds():
-    """Test that unspecified Bounds are accepted and set to None"""
+    """Test that unspecified Bounds are accepted and set to None."""
     template = ProcessTemplate('foo', conditions=[
         (LinkByUID('1', '2'), None),
         [LinkByUID('3', '4'), None],
--- a/gemd/entity/template/tests/test_process_template.py
+++ b/gemd/entity/template/tests/test_process_template.py
@@ -1,9 +1,12 @@
 """Tests of the ProcessTemplate object."""
 import pytest
 
+from gemd.entity.bounds import IntegerBounds
+from gemd.entity.link_by_uid import LinkByUID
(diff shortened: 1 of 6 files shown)


gemd/entity/template/tests/test_process_template.py
```diff
@@ -1,9 +1,12 @@
 """Tests of the ProcessTemplate object."""
 import pytest
 
+from gemd.entity.bounds import IntegerBounds
+from gemd.entity.link_by_uid import LinkByUID
 from gemd.entity.template.process_template import ProcessTemplate
 from gemd.entity.template.condition_template import ConditionTemplate
 from gemd.entity.bounds.real_bounds import RealBounds
+from gemd.json import dumps, loads
```
kroenlein, line 9: Given how you use this below, there is a `copy` method in `gemd.json` you should be using instead.
```suggestion
from gemd.json import copy
```

gemd/entity/template/tests/test_process_template.py
```diff
@@ -39,3 +42,35 @@ def test_allowed_labels():
...
+    template = ProcessTemplate('foo', conditions=[
+        (LinkByUID('1', '2'), None),
+        [LinkByUID('3', '4'), None],
+        LinkByUID('5', '6'),
+        ConditionTemplate('foo', bounds=IntegerBounds(0, 10)),
+    ])
+    assert len(template.conditions) == 4
+    for _, bounds in template.conditions:
+        assert bounds is None
+    copied = loads(dumps(template))
```
kroenlein, line 58:
```suggestion
    copied = copy(template)
```
As above

asantas93: I actually kinda like `loads(dumps(` here because it guarantees full round-trip serialization (which is what I wanted to test), whereas `copy` could conceivably be implemented another way.

kroenlein: I see where you are going.  Acceptable.

PR #110, opened 2020-06-17
Serialization class refactor

kroenlein:
Rebuild of JSON serialization and deserialization for speed (tests now 20 % faster).  Also refactors how deserialization does class resolution to remove explicit class list.


gemd/json/gemd_json.py
```diff
@@ -51,21 +24,23 @@ class GEMDJson(object):
...
-        ProcessRun, MaterialRun, MeasurementRun, IngredientRun,
-        Property, Condition, Parameter, PropertyAndConditions,
-        PropertyTemplate, ConditionTemplate, ParameterTemplate,
-        RealBounds, IntegerBounds, CategoricalBounds, CompositionBounds,
-        MolecularStructureBounds,
-        NominalComposition, EmpiricalFormula,
-        NominalReal, UniformReal, NormalReal, DiscreteCategorical, NominalCategorical,
-        UniformInteger, NominalInteger, Smiles, InChI,
-        FileLink, PerformedSource
+        obj for mod in _modules for _, obj in inspect.getmembers(mod)
```
maxhutch, line 38: I'm not a huge fan of making this implicit: if someone forgets to add a class here we should see it in a test.  But if we do make it implicit, I think it should be heavily documented, e.g. in-line and in the contributing guide and in the serialization docs [here](https://github.com/CitrineInformatics/gemd-python/blob/master/docs/source/depth/serialization.rst).

kroenlein:
Closing, as the automated import proposed here was decided against.

PR #111, opened 2020-06-17
Update objects to require names for all objects

kroenlein:
This PR modifies constructors for all GEMD entities to require a name value and only accepts name as a positional argument.  

The exception to this is IngredientRun, which explicitly removes `name` and `labels` from its argument list since those are not set by users.

### DIFF
--- a/gemd/entity/object/ingredient_run.py
+++ b/gemd/entity/object/ingredient_run.py
@@ -50,7 +50,7 @@ class IngredientRun(BaseObject, HasQuantities):
 
     typ = "ingredient_run"
 
-    def __init__(self, material=None, process=None, mass_fraction=None,
+    def __init__(self, *, material=None, process=None, mass_fraction=None,
                  volume_fraction=None, number_fraction=None, absolute_quantity=None,
                  spec=None, uids=None, tags=None, notes=None, file_links=None):
         BaseObject.__init__(self, name=None, uids=uids, tags=tags,
@@ -77,7 +77,7 @@ class IngredientRun(BaseObject, HasQuantities):
     def name(self, name):
         # This messiness is a consequence of name being an inherited attribute
         if name is not None:
-            warnings.warn("Labels are set implicitly by associating with an "
+            warnings.warn("Name is set implicitly by associating with an "
                           "IngredientSpec; this value will likely be overwritten",
                           DeprecationWarning)
         self.__class__._name_setter(self, name)
@@ -93,7 +93,7 @@ class IngredientRun(BaseObject, HasQuantities):
         return self._labels
 
     @labels.setter
-    @deprecation.deprecated(deprecated_in="0.10", removed_in="0.11",
+    @deprecation.deprecated(deprecated_in="0.12", removed_in="0.13",
                             details="Labels are set implicitly by associating with an "
                                     "IngredientSpec")
     def labels(self, labels):
--- a/gemd/entity/object/ingredient_run.py
+++ b/gemd/entity/object/ingredient_run.py
@@ -139,7 +139,13 @@ class IngredientRun(BaseObject, HasQuantities):
 
     @classmethod
     def from_dict(cls, d):
-        """Suppresses name/label warnings during deserializaton."""
-        with warnings.catch_warnings():
-            warnings.filterwarnings('ignore', category=DeprecationWarning)
-            return super().from_dict(d)
+        # Overload from_dict because we may get name/labels from an independent
+        # resource, but not get the associated IngredientSpec
+        name = d.pop("name", None)
+        labels = d.pop("labels", None)
+        obj = super().from_dict(d)
+        if name is not None:
+            obj._name = name
+        if labels is not None:
+            obj._labels = labels
+        return obj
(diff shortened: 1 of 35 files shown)


kroenlein:
At present, there are incompatibilities with citrine-python, particularly with how it add GEMD-like objects to the deserialization process.

kroenlein:
For @djack: this is blocking https://citrine.atlassian.net/browse/PLA-3337, but if implemented has substantial implications downsteam.

kroenlein:
Note that at present, there are downstream consequences to removing `label` and `name` from the Ingredient constructor.  These arguments have been marked as obsolete, complete with deprecation warnings, for at least 4 months and were inherited in the taurus import.

kroenlein:
This enables:
https://github.com/CitrineInformatics/citrine-python/pull/365/

nextgen-devkit has been forward migrated, I believe, to not be vulnerable to this change:
https://github.com/CitrineInformatics/nextgen-devkit/pull/639

gemd/entity/object/ingredient_run.py
```diff
@@ -53,41 +50,59 @@ class IngredientRun(BaseObject, HasQuantities):
...
+    @property
+    def name(self):
+        """Get name."""
+        return super().name
+
+    @name.setter
+    def name(self, name):
+        # This messiness is a consequence of name being an inherited attribute
+        if name is not None:
+            warnings.warn("Labels are set implicitly by associating with an "
```
bfolie, line 80:
```suggestion
            warnings.warn("Name is set implicitly by associating with an "
```

gemd/entity/object/ingredient_run.py
```diff
@@ -53,41 +50,59 @@ class IngredientRun(BaseObject, HasQuantities):
...
         self.process = process
-        self.labels = labels
-        # this may overwrite name/labels
+        # this will overwrite name/labels if/when they are set
         self.spec = spec
 
+    @property
+    def name(self):
+        """Get name."""
+        return super().name
```
bfolie, line 74: How does this work, given that there are 2 super classes?

gemd/entity/object/ingredient_run.py
```diff
@@ -53,41 +50,59 @@ class IngredientRun(BaseObject, HasQuantities):
...
+        if name is not None:
+            warnings.warn("Labels are set implicitly by associating with an "
+                          "IngredientSpec; this value will likely be overwritten",
+                          DeprecationWarning)
+        self.__class__._name_setter(self, name)
+
+    @classmethod
+    def _name_setter(cls, obj, name):
+        """Misdirection so assignment in from_dict doesn't hit deprecation warning."""
+        super(IngredientRun, cls).name.fset(obj, name)
```
bfolie, line 88: What's this line doing?

gemd/entity/object/tests/test_ingredient_run.py
```diff
@@ -35,3 +35,42 @@ def test_invalid_assignment():
...
+    je = GEMDJson()
+
+    ms_link = LinkByUID(scope='local', id='mat_spec')
+    mr_link = LinkByUID(scope='local', id='mat_run')
+    ps_link = LinkByUID(scope='local', id='pro_spec')
+    pr_link = LinkByUID(scope='local', id='pro_run')
+    spec = IngredientSpec(name='Ingred', labels=['some', 'words'],
+                          process=ps_link, material=ms_link)
+    run = IngredientRun(spec=spec,
+                        process=pr_link, material=mr_link)
```
bfolie, line 57: I think you should add here an assertion that `run.name` and `run.labels` match the `spec` values. Otherwise the test below would pass in a situation where ingredient runs did not have any name/label related information.

bfolie:
If the user gets a serialized ingredient run from a database, and the JSON dictionary includes 'name' and 'labels' fields, as they do now, will that information be properly incorporated without issuing warnings? It looks like the answer is "yes" but I want to make sure I'm reading this correctly

kroenlein:
> If the user gets a serialized ingredient run from a database, and the JSON dictionary includes 'name' and 'labels' fields, as they do now, will that information be properly incorporated without issuing warnings? It looks like the answer is "yes" but I want to make sure I'm reading this correctly

That is the intent and I believe that is the current behavior.  There is only one valid pathway for setting the `name` attribute, which is assigning an IngredientSpec, but if you are reading in a serialized IngredientRun that only has a LinkByUID in `spec`, that information is still more valuable that the other redundant information (e.g., `output_material`).

gemd/entity/object/ingredient_run.py
kroenlein: Dark art that I'm not happy about, but other methods I tried caused problems in libraries that inherited from this.

This expressly sets the name attribute associated with the IngredientRun's parent class (Base Object) without invoking the IngredientRun's overridden setter.  Without this, setting the name via setting an IngredientSpec or deserializing from JSON fires off a deprecation warning every time.

gemd/entity/object/ingredient_run.py
kroenlein: Essentially, because I wanted the setter to issue a deprecation warning, so the getter/setter pair override the parent classes attribute, so I need to reference it explicitly to keep them connected.  Yes, that feels more fragile than I want it to be.

PR #112, opened 2020-06-17
adding uid property to LinkByUID

lkubie:
__repr__ of LinkByUID makes a user think there is a uid property to the object... so I added one. I also left the id property so it is not a breaking change.

### DIFF
--- a/gemd/entity/link_by_uid.py
+++ b/gemd/entity/link_by_uid.py
@@ -52,4 +52,3 @@ class LinkByUID(DictSerializable):
                 entity.add_uid(name, str(uuid.uuid4()))
             scope, uid = next((s, i) for s, i in entity.uids.items())
         return LinkByUID(scope, uid)
-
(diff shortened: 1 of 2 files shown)


lkubie:
Good point, @maxhutch ! Maybe the correct thing here is actually just to change the way the object Prints. It was misleading before.

gemd/entity/link_by_uid.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, scope, id):
         self.id = id
 
     def __repr__(self):
-        return str({"scope": self.scope, "uid": self.id})
+        return str({"scope": self.scope, "id": self.id})
```
maxhutch, line 28: :+1:

maxhutch:
The tests are green; the status just isn't propagating: https://travis-ci.org/github/CitrineInformatics/gemd-python/builds/699472775

PR #114, opened 2020-07-13
Rework cake data example

kroenlein:
This edit of the Cake demo includes several changes reflecting shifting understanding since it was first generated:

* Template scope is now explicitly defined
* A mixer speed Attribute Template is used to create several Process Spec Parameters that have no Run equivalent
* Spec construction is now done via a local method to remove redundant strings
* Many tags and notes were added
* Ingredient names were brought into better alignment with best practice, as well as adding `allowed_names` values to Process Templates
* Fake brand names are added to Material Runs, as opposed to their abstract Spec counterparts
* An argument for defining a FileLink for an image to be attached to the Baked Doneness measurement was added

### DIFF
--- a/gemd/demo/cake.py
+++ b/gemd/demo/cake.py
@@ -122,6 +122,10 @@ def make_cake_templates():
     )
 
     # Objects
+    tmpl["Procuring"] = ProcessTemplate(
+        name="Procuring",
+        description="Buyin' stuff"
+    )
     tmpl["Baking"] = ProcessTemplate(
         name="Baking",
         description='Using heat to promote chemical reactions in a material',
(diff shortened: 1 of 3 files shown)


gemd/demo/cake.py
```diff
@@ -30,6 +30,27 @@
 TEMPLATE_SCOPE = DEMO_SCOPE + '-template'
 
 
+def change_scope(data, *, templates=None):
```
ventura-rivera, line 33: Dynamically changing a global variable seems risky. Why do you need to do this?

kroenlein: This is a minimal change to support having multiple copies of the dataset be able to co-exist on platform.  If the e2e tests are run with CITRINE-NO-PRELOAD=1, for example, this let's us have these tests still meaningful.  This, for example, would have prevented the issues plaguing [nextgen pull 622](https://github.com/CitrineInformatics/nextgen-devkit/pull/622).

ventura-rivera: Gotcha 👍

maxhutch: I agree with @ventura-rivera : wrapping this whole module in a class with scope as one of the class members would be preferable.  We could do that in a fast-follow on though to keep the changes in this PR tighter.

gemd/demo/cake.py
```diff
@@ -916,7 +1063,21 @@ def find_name(name, material):
     run_key = md5.hexdigest()
 
     # Crawl tree and annotate with uids; only add ids if there's nothing there
-    recursive_foreach(cake, lambda obj: obj.uids or obj.add_uid(DEMO_SCOPE, obj.name + run_key))
+    name_count = dict()
+
+    def _disambig(name):
+        nonlocal name_count
```
maxhutch, line 1069: This non-local pattern looks to work, but I think the slightly more pythonic way would be to add `name_count` to the argument list of `_disambig`.

gemd/demo/cake.py
```diff
@@ -639,12 +773,23 @@ def ingredient_kwargs(material):
     )
 
     # Crawl tree and annotate with uids; only add ids if there's nothing there
-    recursive_foreach(cake, lambda obj: obj.uids or obj.add_uid(DEMO_SCOPE, obj.name))
+    seen = set()
+
+    def name_fuzz(obj):
+        # Add fuzz to name in ID as necessary
+        nonlocal seen
```
maxhutch, line 780: This non-local pattern looks to work, but I think the slightly more pythonic way would be to add seen to the argument list of name_fuzz (and maybe a leading underscore).

gemd/demo/cake.py
```diff
@@ -198,31 +233,37 @@ def make_cake_spec(tmpl=None):
...
-            "name": "{} input{}".format(material.name.replace('Abstract ', ''),
-                                        " (Again)" * (count[material.name] - 1)),
+            "name": material.name.lower(),
             "tags": list(material.tags),
             "material": material
         }
 
+    def material_kwargs(material_name, process_tmpl_name, process_kwargs):
+        # Handles some nicities in naming
+        return {
```
maxhutch, line 246: This pattern probably deserves some documentation.  I see what its trying to do, but its not a pattern that I've seen before.

setup.py
```diff
@@ -36,7 +36,7 @@ def run(self):
 
 
 setup(name='gemd',
-      version='0.9.2',
+      version='0.9.3',
```
maxhutch, line 39: `change_scope` is a new public method, so I think this should be a minor version bump.

gemd/demo/cake.py
kroenlein: Migrated to using a closure & generator.

gemd/demo/cake.py
kroenlein: Migrated to using a closure & generator.

gemd/demo/cake.py
kroenlein: Migrated to a wrapper for the constructor (the inverse of this, which I believe is a more common approach)

gemd/demo/cake.py
kroenlein: Wrapping in a class could make sense, but would require generation of a singleton in order to avoid breaking downstream code.

PR #115, opened 2020-07-20
Bugfixes for cake demo

kroenlein:
This PR addresses several issues:
* There were UIDs that collided when compared as lower() strings.  The algorithm for generating UIDs was reworked so that should no longer be any collisions.
* The test suite was modified to catch this.
* There was a linking failure in the ingredient set.  This bug was corrected and is now guarded against through the _make_ingredient sig

### DIFF
--- a/gemd/entity/value/uniform_integer.py
+++ b/gemd/entity/value/uniform_integer.py
@@ -38,7 +38,9 @@ class UniformInteger(IntegerValue):
                 "lower_bound must be an int; got {}({})".format(type(lower_bound), lower_bound))
         if self._upper_bound is not None:
             if lower_bound > self.upper_bound:
-                raise ValueError("lower_bound ({}) must be <= upper_bound ({})".format(lower_bound, self.upper_bound))
+                raise ValueError(
+                    "lower_bound ({}) must be <= upper_bound ({})".format(lower_bound,
+                                                                          self.upper_bound))
         self._lower_bound = int(lower_bound)
 
     @property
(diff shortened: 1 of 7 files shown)


maxhutch:
It looks like several of the tests had been mis-calibrated.  Have you done something else to verify that the result of this example is correct (e.g. upload and render material histories), such that we just have to check for regressions?

gemd/entity/value/uniform_integer.py
```diff
@@ -17,10 +17,35 @@ class UniformInteger(IntegerValue):
...
+    @property
+    def upper_bound(self) -> int:
+        """The upper bound of a uniform distribution."""
+        return int(self._upper_bound)
+
+    @upper_bound.setter
+    def upper_bound(self, upper_bound: int) -> None:
+        """The upper bound of a uniform distribution."""
+        # This check is necessary to handle JSON serialization behavior under 3.5
+        assert float(int(upper_bound)) == float(upper_bound), "upper bound must be an int"
```
maxhutch, line 50: I think if `upper_bound` is a string, this will throw a ValueError due to `int("foo")` rather than an assertion error and not return this error message.

kroenlein: That is correct.  But you think a ValueError would be more appropriate here than a TypeError?

maxhutch: I care most about the message and second most about consistency.  No strong opinions about the type of error.

PR #119, opened 2020-09-02
Fixed minor data issues in cake demo

kroenlein:
Bug fixes for the cake demo:
* Typos
* There were no allowed_names lists
* Some objects were shadowed
* A URL needed updating

There are no changes to interfaces and all existing tests continue to pass.

### DIFF
--- a/gemd/demo/cake.py
+++ b/gemd/demo/cake.py
@@ -334,7 +334,7 @@ def make_cake_spec(tmpl=None):
             "conditions": [
                 Condition(name='Cooking time',
                           template=tmpl['Cooking time'],
-                          origin=Origin.SPECIFIED.value,
+                          origin=Origin.SPECIFIED,
                           value=NormalReal(mean=50, std=5, units='min'))
             ],
             "parameters": [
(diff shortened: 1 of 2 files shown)


kroenlein:
> I'm surprised you need the `.value` for the origins. I would expect it to work w/o, and I would expect common usage to omit the `.value`.

It worked without it, but the linter was complaining about the fact that I was passing an Origin instead of a string.  Happy to revert.

maxhutch:
> > I'm surprised you need the `.value` for the origins. I would expect it to work w/o, and I would expect common usage to omit the `.value`.
> 
> It worked without it, but the linter was complaining about the fact that I was passing an Origin instead of a string. Happy to revert.

I think the right thing to satisfy the linter is to define the argument to the setter as a Union[Origin, str], which is value add.

kroenlein:
> > > I'm surprised you need the `.value` for the origins. I would expect it to work w/o, and I would expect common usage to omit the `.value`.
> > 
> > 
> > It worked without it, but the linter was complaining about the fact that I was passing an Origin instead of a string. Happy to revert.
> 
> I think the right thing to satisfy the linter is to define the argument to the setter as a Union[Origin, str], which is value add.

Yes; that literally occurred to me after I hit `Comment` and I was in the middle of the edit.

kroenlein:
I've modified this to not explicitly invoke .value on the Origin objects.  I took a look at augmenting everything with typehints, but that was much larger than this PR motivates.

PR #121, opened 2020-09-11
Make CaseInsensitiveDict delete obsolete keys

kroenlein:
This is a bug fix to address issues where a key would be remembered by CaseInsensitiveDict even after the key had been deleted from the dictionary.  It generally follows the principle that a set operation requires using the exact key whereas a get or a delete should be forgiving of case inconsistencies.

### DIFF
--- a/gemd/entity/case_insensitive_dict.py
+++ b/gemd/entity/case_insensitive_dict.py
@@ -108,8 +108,6 @@ class CaseInsensitiveDict(dict):
         """
         Remove and return a (key, value) pair from the dictionary.
 
-        Pairs are returned in LIFO order.
-
         popitem() is useful to destructively iterate over a dictionary, as often used
         in set algorithms.  If the dictionary is empty, calling popitem() raises a
         KeyError.
--- a/gemd/entity/case_insensitive_dict.py
+++ b/gemd/entity/case_insensitive_dict.py
@@ -159,6 +159,9 @@ class CaseInsensitiveDict(dict):
         """
         if mapping is None:
             mapping = dict()
+            no_mapping = True
+        else:
+            no_mapping = False
         for key in list(mapping.keys()) + list(kwargs.keys()):
             if key.lower() in self.lowercase_dict:
                 prev = self.lowercase_dict[key.lower()]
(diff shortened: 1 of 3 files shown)


gemd/entity/case_insensitive_dict.py
```diff
@@ -50,14 +55,128 @@ def get(self, key: str, default=None):
...
+            The result to return if the key is not present.
+
+        Returns
+        -------
+        Any
+            The value associated with the case-insensitive version of `key`, or None
+            if `key` is not present.
+
+        """
+        if default is _RaiseKeyError:
```
bfolie, line 97: Why not make the default `None`?

gemd/entity/case_insensitive_dict.py
```diff
@@ -50,14 +55,128 @@ def get(self, key: str, default=None):
...
+            val = super().pop(self.lowercase_dict.get(key.lower()), default)
+        if key in self:
+            del self.lowercase_dict[key.lower()]
+        return val
+
+    def popitem(self) -> Tuple:
+        """
+        Remove and return a (key, value) pair from the dictionary.
+
+        Pairs are returned in LIFO order.
```
bfolie, line 111: Since we support python versions older than 3.7, isn't this dependent on which version of python the user has?

gemd/entity/case_insensitive_dict.py
kroenlein: Because `None` is a legitimate value.  I need to catch that the argument was not passed.

gemd/entity/case_insensitive_dict.py
kroenlein: I was a little lazy in copy/paste of the docs.  I can clean this up if desired.

gemd/entity/case_insensitive_dict.py
bfolie: Yes, I understand now

gemd/entity/tests/test_case_insensitive_dict.py
```diff
@@ -56,3 +56,51 @@ def test_contains():
         assert k in data_dict
 
     assert 'not_a_key' not in data_dict
+
+
+def test_full_api():
```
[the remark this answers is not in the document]
kroenlein: Test name changed, comments to indicate what's going on added.

gemd/entity/case_insensitive_dict.py
bfolie: I think just removing this line would suffice, since lines 117-118 are strictly more informative.

bfolie (approved):

PR #122, opened 2020-09-16
Augment set of acceptable units

kroenlein:
As per https://citrine.atlassian.net/browse/PLA-5662, I've augmented the set of acceptable unit names to allow unit definition via capitalized unit names, e.g., Celcius.  Specifying units in this way is non-standard and runs against best practice, but that doesn't mean that we shouldn't accept these given that there is no ambiguity in definition.


gregor-robinson (approved):
LGTM

PR #123, opened 2020-09-18
Modify Strehlow & Cook to have heterogeneous units

kroenlein:
To improve the value of Strehlow and Cook in a test context, this arbitrarily maps all band gaps with an uncertainty statement into units of attojoules.  Thus, two different units are present in the resultant corpus for the same property.

### DIFF
--- a/gemd/demo/strehlow_and_cook.py
+++ b/gemd/demo/strehlow_and_cook.py
@@ -28,6 +28,7 @@ from gemd.entity.bounds.real_bounds import RealBounds
 
 from gemd.enumeration.origin import Origin
 
+from gemd.units import convert_units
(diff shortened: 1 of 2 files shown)


maxhutch (commented):
Do we want to only do this for some of them so that the dataset has mixed units?

kroenlein:
> Do we want to only do this for some of them so that the dataset has mixed units?

This does yield mixed units -- not all band gaps have uncertainties.

maxhutch:
> This does yield mixed units -- not all band gaps have uncertainties.

Ahh, I didn't catch that.  Great!

maxhutch (approved):

PR #124, opened 2020-09-23
Feature/change units

kroenlein:
Adding an explicit interface for changing the target units file for gemd-python.  

Most of these changes revolve around bringing the units conversion file in-line with Contributing code standards.  The private method `_unit_to_str` has also been removed as it was not used elsewhere in the module and the only effect was testing an exception that is impossible with how it is called.

### DIFF
--- a/gemd/units/tests/test_parser.py
+++ b/gemd/units/tests/test_parser.py
@@ -44,7 +44,7 @@ def test_parse_none():
 
 
 @contextmanager
-def change_units(filename):
+def _change_units(filename):
     try:
         change_definitions_file(filename)
         yield
@@ -57,7 +57,8 @@ def test_file_change():
     assert convert_units(1, 'm', 'cm') == 100
     with pytest.raises(UndefinedUnitError):
         assert convert_units(1, 'usd', 'usd') == 1
-    with change_units(filename=pkg_resources.resource_filename("gemd.units", "tests/test_units.txt")):
+    with _change_units(filename=pkg_resources.resource_filename("gemd.units",
+                                                                "tests/test_units.txt")):
         with pytest.raises(UndefinedUnitError):
             assert convert_units(1, 'm', 'cm') == 100
         assert convert_units(1, 'usd', 'usd') == 1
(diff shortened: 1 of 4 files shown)


gemd/units/tests/test_parser.py
```diff
@@ -39,3 +40,19 @@ def test_parse_unexpected():
...
+    assert convert_units(1, 'm', 'cm') == 100
+    with pytest.raises(UndefinedUnitError):
+        assert convert_units(1, 'usd', 'usd') == 1
+    change_definitions_file(
+        filename=pkg_resources.resource_filename("gemd.units", "tests/test_units.txt")
+    )
+    with pytest.raises(UndefinedUnitError):
+        assert convert_units(1, 'm', 'cm') == 100
+    assert convert_units(1, 'usd', 'usd') == 1
+    change_definitions_file(
```
maxhutch, line 56: If this change back is to make sure this test state doesn't pollute other tests, then it would be best to wrap the rest of the test in a `try... except ... finally` block and stick the convert back into the `finally`.  Even better would be to use a context manager, which provides the same behavior but without relying on `try... except`

setup.py
```diff
@@ -49,6 +49,7 @@ def run(self):
               'demo/toothpick.jpg',
               'units/citrine_en.txt',
               'units/constants_en.txt'
+              'units/tests/test_units.txt'
```
maxhutch, line 52: Do we need to include this in the package?  The tests sources aren't included (I don't think), so the file is not referenced by anything in the package (I think).

maxhutch (changes requested):
The feature logic looks great.

I made two inline comments around managing the test state.

setup.py
kroenlein: When I neglected to include it in the package, Jenkins failed this test because it couldn't find the file.

gemd/units/tests/test_parser.py
kroenlein: Good catch

setup.py
maxhutch: Ahh, yeah, and I don't see a test_package option or anything.  :+1:

maxhutch (approved):

PR #125, opened 2020-10-15
Added mechanism for checking consistency between bounds and values

kroenlein:
In order to support SDK work around template generation and local validations, this PR provides a mechanism for testing consistency between Values and Bounds.

The explicit calls `to_bounds` seems a little less elegant than implicitly supporting them in the Bounds `contains` methods, but that would have required a little more iter-object awareness than seemed appropriate without discussion.  If that seems like the more prudent route, ading the appropriate methods would be straight forward.

### DIFF
--- a/gemd/entity/value/base_value.py
+++ b/gemd/entity/value/base_value.py
@@ -28,4 +28,3 @@ class BaseValue(DictSerializable):
             :class:`bounds <gemd.entity.bounds.base_bounds.BaseBounds>`.
 
         """
-        pass
--- a/gemd/entity/value/base_value.py
+++ b/gemd/entity/value/base_value.py
@@ -17,7 +17,7 @@ class BaseValue(DictSerializable):
     typ = "value"
 
     @abstractmethod
-    def to_bounds(self) -> BaseBounds:
+    def _to_bounds(self) -> BaseBounds:
         """
         Return the smallest bounds object that is consistent with the Value.
 
--- a/gemd/entity/value/base_value.py
+++ b/gemd/entity/value/base_value.py
@@ -1,5 +1,8 @@
 """Base class for all values."""
 from gemd.entity.dict_serializable import DictSerializable
+from gemd.entity.bounds.base_bounds import BaseBounds
+
+from abc import abstractmethod
 
 
 class BaseValue(DictSerializable):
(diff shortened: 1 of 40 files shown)


gemd/entity/value/base_value.py
```diff
@@ -12,3 +15,16 @@ class BaseValue(DictSerializable):
     """
 
     typ = "value"
+
+    @abstractmethod
+    def to_bounds(self) -> BaseBounds:
```
maxhutch, line 20: Is there a use case for this method outside of `contains`?  Should this method be "protected"?

gemd/entity/value/categorical_value.py
```diff
@@ -8,3 +11,16 @@ class CategoricalValue(BaseValue):
 
     All category names must be in unicode.
     """
+
+    @abstractmethod
+    def to_bounds(self) -> CategoricalBounds:
```
maxhutch, line 16: Non-blocking comment: I think you can get this typing behavior by using a generic in `BaseValue` that would avoid needing to re-define the whole method and docstring.  Here's an example: https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/resources/data_concepts.py#L212

maxhutch (commented):
Logic looks good.  One question about the `to_bounds` method and another about typing, both inline.

gemd/entity/value/base_value.py
kroenlein: The obvious use case would be building the union of two bounds.  It's not obvious that this is the cleanest implementation, but a method call like
```
full_range = value1.to_bounds.union(value2.to_bounds)
```
might make sense (unattractive sig, yes).  In particular, 
```
full_range = value1.union(value2)
```
probably _wouldn't_ make sense.  Another viable option would be
```
full_range = BaseBounds(value1).union(value)
```
but the inheritance gets a little confused (i.e., I called a static method on CategoricalBounds but it returned a RealBounds).  

This last scenario would allow changing this to a protected method.

maxhutch: I'd be inclined to make it protected until `union` exists.  Its easy to make a protected method public than to make a public method protected.

gemd/entity/value/categorical_value.py
kroenlein: I might be missing something, but I don't think that does quite what this is trying to do.  Happy to chat if you'd like to follow up.

maxhutch (approved):
LGTM

PR #129, opened 2021-02-18
Added additional support for capitalized units

kroenlein:
As per PLA-5662, I am further increasing the number of capitalized unit names recognized by our definition file.


maxhutch (changes requested):
Minimally, we should add some test coverage of the added units.  That should be straight-forward in the `test_parser.py` file.

Pint added a case insensitive option in a recent release (see https://github.com/hgrecco/pint/pull/1147, which is included in https://github.com/hgrecco/pint/releases/tag/0.16.1 I think).  Would that be a cleaner option?

gemd/units/citrine_en.txt
```diff
@@ -168,114 +168,114 @@ count = []
...
-decade = 10 * year
-century = 100 * year = _ = centuries
-millennium = 1e3 * year = _ = millennia
-eon = 1e9 * year
-shake = 1e-8 * second
+minute = 60 * second = min = Minute
+hour = 60 * minute = hr = Hour
+day = 24 * hour = d = Day
+week = 7 * day = Week
+fortnight = 2 * week = Fortnight
```
maxhutch, line 218: :man_dancing:

maxhutch (approved):
ouch

PR #130, opened 2021-02-19
Dropping python 3.5 support

maxhutch:
Python 3.5 is end-of-life.  Users should upgrade
to Python 3.6+


kroenlein (approved):
Makes sense, given that EOL on python 3.5 was 6 months ago.  Changes look good.

PR #131, opened 2021-03-26
add util function to make the index 

lkubie:
add function to make the index used in substitute_links and substitute_objects methods


gemd/util/impl.py
```diff
@@ -69,6 +69,18 @@ def _substitute(thing,
...
+    """
+    Given a list of GEMD objects, generates an index that can be used for substitute_links and substitute_objects methods.
+
+    :gems list: List of GEMD objects from which to create an index
+
+    """
+    gem_index = {}
+    for gem in gems:
+        for k, v in gem.uids.items():
+            gem_index[(k,v)] = gem
```
kroenlein, line 82: Should we be invoking `set_uuids` here to guarantee that every object is mappable?

kroenlein (commented):
Question for consideration.

gemd/util/impl.py
lkubie: I don't think we should. A person would never need an object within the index to be rehydrated if it never had a unique id; there would have been no scope/id pair to form the LinkByUID Object pair in the first place. Additionally, users use the "id" scoped id, traditionally, to keep track on registered versus non-registered objects.

PR #132, opened 2021-03-26
codeowners

djack:
same codeowners group that we are using for gemd-docs for now.


kroenlein (approved):
Subject to the one-or-two-groups question asked in gemd-docs.

kroenlein (approved):
LGTM

PR #135, opened 2021-03-29
Add util function to make the substitution index

kroenlein:
Added a function to make the index used in the substitute_objects method.

Follow-up to #131.

### DIFF
--- a/gemd/util/impl.py
+++ b/gemd/util/impl.py
@@ -70,18 +70,24 @@ def _substitute(thing,
     return new
 
 
-def make_index(gems):
+def make_index(obj):
     """
     Generates an index that can be used for substitute_links and substitute_objects methods.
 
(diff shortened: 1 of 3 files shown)


lkubie (commented):
Please use a more descriptive variable name for make_index(). The variable name Obj does not make clear that this is a list of objects.

In the docstring, please be more expressive as to what this function can accept.

kroenlein:
> Please use a more descriptive variable name for make_index(). The variable name Obj does not make clear that this can accept one of several data types.
> 
> In the docstring, please be more expressive as to what this function can accept. This will allow users to use the method with more confidence

The docstring is now more expressive about the potential argument types.  

The variable name for make_index is still `obj` since that's consistent with the other methods in the file.  A more correctly scoped word for `a wide range of types` than the generic "object" is not clear, but suggestions would be welcome.

PR #137, opened 2021-04-16
Move secret tokens to web UI

maxhutch:
Travis has improved its handling of secret tokens and its easier to track them in the travis UI now.  This removes them from the travis file so they can be dynamically managed there.


kroenlein (approved):
LGTM

PR #141, opened 2021-05-20
Remove `mas` from unit abbreviations

kroenlein:
As per PLA-7408, `milliarcseconds` had the abbreviation `mas`.  When combined with `parse_units` being insensitive to trailing `%` or `s` (allowing `miles` to match `mile`), this meant that the unit `mass %` was being interpreted as `milliarcseconds`.  As an easy fix, we are removing `mas` from the synonym list.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 
 setup(name='gemd',
-      version='0.17.8',
+      version='0.17.9',
       url='http://github.com/CitrineInformatics/gemd-python',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -30,4 +30,4 @@
           'Programming Language :: Python :: 3.7',
           'Programming Language :: Python :: 3.8',
       ],
-)
+      )
```
kroenlein, line 33: This indent is to satisfy flake8.

bfolie (approved):

PR #142, opened 2021-05-30
Feature/remove recursion

kroenlein:
To increase performance, `recursive_flatmap` and `recursive_foreach` are now queues rather than recursive subroutines.

As well, a caching isinstance generator was created in order to improve introspection speed.

Both were identified as bottlenecks following profiling work in citrine-e2e-examples.


bfolie:
Do you have any profiling data? That is, timing on test cases before and after these changes?

kroenlein:
> Do you have any profiling data? That is, timing on test cases before and after these changes?

The test suite for this library, on my machine, executes in 5.2 seconds, down from 6.8 prior to this commit.  The bulk of that is the test_sac call, which is currently dominated by calls into json.dumps & json.loads.

The e2e2 perfumes example had 84 seconds associated with recursive_foreach and subcalls, 32 seconds of which was isinstance or subcalls.  The shift from recursion to a queue reduced recursive_foreach to 36 tottime (- 5 seconds) and isinstance to 5.6 s cumulative time (-26 seconds).

bfolie (approved):

PR #143, opened 2021-05-30
Made scope argument to flatten optional

kroenlein:
Modified `scope` argument to `flatten` to be optional, so that if a set of objects already have a full set of uids, there is no need to pass a dummy argument.

Flattening a graph that is not fully annotated with UIDs raises a ValueError.


scripts/validate-version-bump.sh
```diff
@@ -15,7 +15,7 @@ extract_minor() {
 }
 
 extract_patch() {
-  sed -En "s/.*([0-9])/\1/p" <<< "$1"
+  sed -En "s/.*\.([0-9]+)/\1/p" <<< "$1"
```
bfolie, line 18: good catch. I should transfer this fix to citrine-python

bfolie (approved):

PR #148, opened 2021-07-23
Add inch-pounds to default unit set

kroenlein:
As per [PLA-7928](https://citrine.atlassian.net/browse/PLA-7928), adding inch-pounds as a unit for torque.


maxhutch:
Can we include a test that converts between inch-pound and foot-pound?

maxhutch (approved):

PR #149, opened 2021-09-23
Migrate arguments for from_entity

kroenlein:
In light of 
* the policy around positional arguments
* the original design choice of including the `name` argument to create the `scope` attribute of a LinkByUID
* the lack of documentation of how the `from_entity` method works

this PR adds a `scope` argument and deprecates both the use of a positional argument and use of the `name` keyword for building a `LinkByUID.from_entity()`.

Unanswered here is if there should be a clever means for changing default scope from `auto` to `id` in the case where gemd-python has been imported by citrine-python.  Historically, the answer was no to avoid the surprising behavior of an import changing a behavior; but I do not believe that was discussed in depth.

### DIFF
--- a/gemd/entity/link_by_uid.py
+++ b/gemd/entity/link_by_uid.py
@@ -1,5 +1,6 @@
 """A unique id that stands in for a data object."""
 import uuid
+from warnings import warn
 
 from gemd.entity.dict_serializable import DictSerializable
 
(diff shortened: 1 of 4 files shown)


gemd/entity/link_by_uid.py
```diff
@@ -28,27 +29,46 @@ def __repr__(self):
...
+        if name is None:
+            if scope is None:
+                scope = "auto"  # set default
+        elif scope is None:
+            warn("The positional argument 'name' is deprecated.  When selecting a default scope, "
+                 "use the 'scope' keyword argument.", DeprecationWarning)
+            scope = name
+        else:
+            if scope is not None:
+                raise ValueError("Specify the 'name' parameter or 'scope' parameter, not both.")
```
djack, line 66: I don't actually think there's a logic error here, but this code took me an absurdly long time to parse for correctness. Could we emulate a pattern matching style with something like 
```suggestion
        if 
        if name is None and scope is None:               # don't break backwards compatibility when the user does nothing
            scope = "auto"
        elif name is not None and scope is None:      # deprecation warning existing users using only the "name" param
            warn("The positional argument 'name' is deprecated.  When selecting a default scope, "
                 "use the 'scope' keyword argument.", DeprecationWarning)
            scope = name
        elif name is not None and scope is not None:  
                raise ValueError("Specify the 'name' parameter or 'scope' parameter, not both.")
```

djack (dismissed):

gemd/entity/link_by_uid.py
kroenlein: Typo in the above, but code following this intent is now pushed.

djack (approved):

PR #150, opened 2021-09-27
Equals

kroenlein:
As per [PLA-8254](https://citrine.atlassian.net/browse/PLA-8254), this implements GEMD object comparison, crawling over an entire material history and respecting the relationship between LinkByUIDs and the Base Entities they reference.

### DIFF
--- a/gemd/entity/base_entity.py
+++ b/gemd/entity/base_entity.py
@@ -70,3 +70,18 @@ class BaseEntity(DictSerializable):
 
         """
         self.uids[scope] = uid
+
+    # Note that this could violate transitivity -- Link(scope1) == obj == Link(scope2)
+    def __eq__(self, other):
+        from gemd.entity.link_by_uid import LinkByUID
+        if isinstance(other, LinkByUID):
+            return self.uids.get(other.scope) == other.id
+        else:
+            result = super().__eq__(other)
+            return result
+
+    # Note the hash function checks if objects are identical, as opposed to the equals method,
+    # which checks if fields are equal.  This is because BaseEntities are fundamentally
+    # mutable objects.
+    def __hash__(self):
+        return super().__hash__()
--- a/gemd/entity/base_entity.py
+++ b/gemd/entity/base_entity.py
@@ -82,6 +82,8 @@ class BaseEntity(DictSerializable):
 
     # Note the hash function checks if objects are identical, as opposed to the equals method,
     # which checks if fields are equal.  This is because BaseEntities are fundamentally
-    # mutable objects.
+    # mutable objects.  Note that if you define an __eq__ method without defining a __hash__
+    # method, the object will become unhashable.
+    # https://docs.python.org/3/reference/datamodel.html#object.__hash
     def __hash__(self):
         return super().__hash__()
--- a/gemd/entity/object/process_run.py
+++ b/gemd/entity/object/process_run.py
@@ -121,6 +121,9 @@ class ProcessRun(BaseObject, HasConditions, HasParameters, HasSource):
             if result is True and isinstance(other, ProcessRun):
                 if len(self.ingredients) == len(other.ingredients):
                     result = all(ing in other.ingredients for ing in self.ingredients)
+                elif (not self.ingredients and self.uids) \
+                        or (not other.ingredients and other.uids):
+                    result = True  # One can be empty if you flattened
                 else:
                     result = False
         finally:
@@ -130,6 +133,8 @@ class ProcessRun(BaseObject, HasConditions, HasParameters, HasSource):
 
     # Note the hash function checks if objects are identical, as opposed to the equals method,
     # which checks if fields are equal.  This is because BaseEntities are fundamentally
-    # mutable objects.
+    # mutable objects.  Note that if you define an __eq__ method without defining a __hash__
+    # method, the object will become unhashable.
+    # https://docs.python.org/3/reference/datamodel.html#object.__hash
     def __hash__(self):
         return super().__hash__()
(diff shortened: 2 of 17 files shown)


gemd/entity/object/process_spec.py
```diff
@@ -56,31 +57,53 @@ class ProcessSpec(BaseObject, HasParameters, HasConditions, HasTemplate):
...
-        if ingred in self._ingredients:
-            self._ingredients.remove(ingred)
-
     @property
     def output_material(self):
         """Get the output material spec."""
         return self._output_material
+
+    # To avoid infinite recursion, fast return on revisit
+    global eq_seen
```
maxhutch, line 86: Is this just a global because you can't change the signature of `__eq__`?  Why put the global in the class instead of outside of it?

gemd/entity/object/process_spec.py
```diff
@@ -56,31 +57,53 @@ class ProcessSpec(BaseObject, HasParameters, HasConditions, HasTemplate):
...
+        result = super().__eq__(other)
+
+        # Equals needs to crawl into ingredients
+        if result is True and isinstance(other, ProcessSpec):
+            if len(self.ingredients) == len(other.ingredients):
+                result = all(ing in other.ingredients for ing in self.ingredients)
+            else:
+                result = False
+
+        eq_seen.remove((self, other))
```
maxhutch, line 102: This might be missed if an exception is raised between the `equ_seen.add` and this.  I suggest using a try/catch/finally to make sure its removed.

gemd/entity/base_entity.py
```diff
@@ -70,3 +70,18 @@ def add_uid(self, scope, uid):
...
+        if isinstance(other, LinkByUID):
+            return self.uids.get(other.scope) == other.id
+        else:
+            result = super().__eq__(other)
+            return result
+
+    # Note the hash function checks if objects are identical, as opposed to the equals method,
+    # which checks if fields are equal.  This is because BaseEntities are fundamentally
+    # mutable objects.
+    def __hash__(self):
```
maxhutch, line 86: What's the purpose of this definition?  Doesn't it automatically inherit from its parent?

gemd/entity/object/process_spec.py
```diff
@@ -56,31 +57,53 @@ class ProcessSpec(BaseObject, HasParameters, HasConditions, HasTemplate):
...
         """Get the output material spec."""
         return self._output_material
+
+    # To avoid infinite recursion, fast return on revisit
+    global eq_seen
+    eq_seen = set()
+
+    def __eq__(self, other):
+        if (self, other) in eq_seen:  # Cycle encountered
+            return True  # This will functionally be & with the correct result of ==
```
maxhutch, line 91: Does this get lit up in the tests?

maxhutch (commented):
Overall, this looks like a reasonable approach.  It'd be nice to get a bit more test coverage of the material history crawling logic; the unit tests added all look very simple.  A few questions inline.

gemd/entity/object/process_spec.py
kroenlein: I'd initially skipped the try-catch because of performance concerns, but it seems like this would be prudent.

gemd/entity/object/process_spec.py
kroenlein: Yes; you can't change the sig of __eq__ and flow of control moves between several modules / is re-entrant.  I wish there was a way to get a tighter scope, but a closure wouldn't get ahead of it.  It's possible there's a better thread/exception safe approach, but I was also trying to maintain velocity.

Not opposed to moving the declaration to the top of file, but wrote it this way to be close to where it is used in the file.

gemd/entity/base_entity.py
kroenlein: It actually does not; you need to explicitly define a __hash__ function every time you define an __eq__.  I went into this expecting ^^ would have been correct.

gemd/entity/object/process_spec.py
kroenlein: Yes.  When the Process Spec iterates over its Ingredient Specs, each `ingredient.process` bounces through this check.

gemd/entity/object/process_spec.py
kroenlein: Better approach implemented, using the object itself as the state vector.

gemd/entity/base_entity.py
kroenlein: https://docs.python.org/3/reference/datamodel.html#object.__hash__

maxhutch: That might be worth mentioning in the inline comments

maxhutch (approved):
LGTM.  It might benefit from a bit more developer docs in the comments.

gemd/entity/object/process_run.py
```diff
@@ -121,6 +121,9 @@ def __eq__(self, other):
             if result is True and isinstance(other, ProcessRun):
                 if len(self.ingredients) == len(other.ingredients):
                     result = all(ing in other.ingredients for ing in self.ingredients)
+                elif (not self.ingredients and self.uids) \
+                        or (not other.ingredients and other.uids):
+                    result = True  # One can be empty if you flattened
```
maxhutch, line 126: So they are equal if they have matching ingredients OR one of them doesn't have any ingredients defined?

kroenlein: Mostly correct.  If one have them has no ingredients and there was a UID dictionary to validate.  

Intended to compensate for the fact that `ingredients` is set implicitly by a relationship that wouldn't be initialized if you just compared the result of `dataset.process_runs.get()` in citrine-python.

kroenlein: Those aren't compared here - they are tested in `DictSerializable.__eq__`.  I'm intending to make sure that _any_ uid exists.  I would be open to changing that line to (len() == 0 / len() != 0) if that reads better.

maxhutch: Yeah, that's what I meant.

maxhutch (approved):

PR #153, opened 2021-10-19
Non-null allowed_names and _labels

kroenlein:
[The docs](https://citrineinformatics.github.io/gemd-docs/specification/object-templates/#process-template) specify that the default state for `allowed_names` and `allowed_labels` for Process Templates is supposed to be the empty set, rather than an explicit None.  This PR brings this behavior into line with the docs and with the backend implementation.

### DIFF
--- a/gemd/entity/template/process_template.py
+++ b/gemd/entity/template/process_template.py
@@ -72,9 +72,8 @@ class ProcessTemplate(BaseTemplate, HasConditionTemplates, HasParameterTemplates
 
     @allowed_names.setter
     def allowed_names(self, allowed_names):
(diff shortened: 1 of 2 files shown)


kroenlein:
> Isn't there a third option? Continue to define empty list as unconstrained, but change the backend behavior so that it doesn't mutate the `None` to an empty list?

That would be the first step in in option 2.  It would put us in a position where we definitely have a backend migration, as opposed to where we have one in one scenario and don't in another.  It still leaves no functional distinction between the two states, which (if this is the final state) raises the question of why have two ways of expressing the same thing.

PR #155, opened 2021-10-20
Add all_dependencies method to Base Entities

kroenlein:
This is an initial effort to migrate [_all_dependencies](https://github.com/CitrineInformatics/citrine-python/blob/d164744c99b17f935a09935c2e57d2f056675477/src/citrine/_utils/batcher.py#L120) into gemd-python.  At present, recursive_foreach and recursive_flatmap can crawl a structure to recover all objects but cannot do so in a way that reveals layers that are helpful in batching. 

The current structure of the method calls don't feel quite right, and at least feel like they motivate creation of a has_spec mixin.

### DIFF
--- a/gemd/entity/base_entity.py
+++ b/gemd/entity/base_entity.py
@@ -1,8 +1,8 @@
 """Base class for all entities."""
-from abc import abstractmethod
-from typing import Optional, Iterable, List, Set, FrozenSet, Mapping, Dict
+from typing import Optional, Union, Iterable, List, Set, FrozenSet, Mapping, Dict
 
 from gemd.entity.dict_serializable import DictSerializable
+from gemd.entity.has_dependencies import HasDependencies
 from gemd.entity.case_insensitive_dict import CaseInsensitiveDict
 from gemd.entity.setters import validate_list
 
@@ -102,9 +102,17 @@ class BaseEntity(DictSerializable):
 
         return LinkByUID(scope=scope, id=uid)
 
-    @abstractmethod
-    def all_dependencies(self) -> Set['BaseEntity']:
+    def all_dependencies(self) -> Set[Union["BaseEntity", "LinkByUID"]]:
         """Return a set of all immediate dependencies (no recursion)."""
+        result = set()
+        queue = [type(self)]
+        while queue:
+            cls = queue.pop()
+            if issubclass(cls, HasDependencies) and \
+                    "_local_dependencies" not in cls.__abstractmethods__:
+                result |= cls._local_dependencies(self)
+                queue.extend(cls.__bases__)
+        return result
 
     @staticmethod
     def _cached_equals(this: 'BaseEntity',
--- a/gemd/entity/base_entity.py
+++ b/gemd/entity/base_entity.py
@@ -108,7 +108,7 @@ class BaseEntity(DictSerializable):
         return LinkByUID(scope=scope, id=uid)
 
     @abstractmethod
-    def all_dependences(self) -> 'Set[BaseEntity]':
+    def all_dependencies(self) -> 'Set[BaseEntity]':
         """Return a set of all immediate dependencies (no recursion)."""
 
     @staticmethod
--- a/gemd/entity/base_entity.py
+++ b/gemd/entity/base_entity.py
@@ -1,5 +1,6 @@
 """Base class for all entities."""
-from typing import Optional
+from abc import abstractmethod
+from typing import Optional, Set
 
 from gemd.entity.dict_serializable import DictSerializable
 from gemd.entity.case_insensitive_dict import CaseInsensitiveDict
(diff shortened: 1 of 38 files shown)


gemd/entity/base_entity.py
```diff
@@ -105,6 +106,10 @@ def to_link(self,
 
         return LinkByUID(scope=scope, id=uid)
 
+    @abstractmethod
+    def all_dependences(self) -> 'Set[BaseEntity]':
```
sfriedowitz, line 110: Is this a typo? I think this method should be `all_dependencies`.

gemd/entity/object/base_object.py
```diff
@@ -54,3 +54,50 @@ def file_links(self):
     @file_links.setter
     def file_links(self, file_links):
         self._file_links = validate_list(file_links, FileLink)
+
+    def all_dependences(self):
+        """Return a set of all immediate dependencies (no recursion)."""
```
sfriedowitz, line 59: Why no recursion? Does method you added to citrine_python traverse the whole graph or just adjacent nodes?

sfriedowitz: Since GEMD objects are now comparable by equality, couldn't you just call `.all_dependencies` on each object you find since this method exists on all entity types. And then `set()` the outputs to remove any duplicates?

kroenlein: The intent is to only return one layer of dependencies.  This isn't as simple as filtering the values of an as_dict because, for example, attributes may contain templates, and that requires navigating 2 layers of objects.  On the other hand, you don't want to include the spec.template object.

kroenlein:
> I agree with this for `HasSpec`. I have run into this issue before, and it would be potentially nice to have a more structured inheritance pattern there to distinguish between run and spec objects.
> 
> Would it be alot of work?

I started prototyping it and it turned into a research project about overloading python setters because ingredient_run has special logic.

gemd/entity/object/base_object.py
```diff
@@ -52,19 +60,52 @@ def _attribute_has_setter(cls, name: str) -> bool:
...
+                        result.add(condition.template)
+
+        if isinstance(self, (IngredientRun, IngredientSpec, MeasurementRun)):
+            if self.material is not None:
+                result.add(self.material)
+        if isinstance(self, (IngredientRun, IngredientSpec, MaterialRun, MaterialSpec)):
+            if self.process is not None:
+                result.add(self.process)
+
+        return result
```
bfolie, line 111: This logic feels wrong because it requires `base_object` to know about the structure of its child classes.

gemd/entity/object/base_object.py
```diff
@@ -52,19 +60,52 @@ def _attribute_has_setter(cls, name: str) -> bool:
...
+        from gemd.entity.object.has_template import HasTemplate
+
+        from gemd.entity.object import IngredientRun, MaterialRun, MeasurementRun  # no ProcessRun
+        from gemd.entity.object import IngredientSpec, MaterialSpec  # no ProcessSpec
+
+        result = set()
+
+        for typ in (HasParameters, HasConditions, HasProperties, HasSpec, HasTemplate):
+            if isinstance(self, typ):
+                result |= typ.all_dependencies(self)
```
bfolie, line 95: Why use the in-place union operation here, but `result.add` below? They seem to me like they both do the same thing.

kroenlein: `add` adds a single element to the set, where as `|=` adds all elements of a second set (like `append` & `extend` for lists).  If you'd prefer all method calls, `result |=` could be changed to `result.update()`

gemd/entity/object/base_object.py
kroenlein: Yeah, I agree that the `isinstance(self, GEMD)` would look nicer if we could test mixins instead.  No obvious solution for `MaterialSpec` since `PropertyAndConditions` is just structured differently than `MeasurementRun`'s property.  A `HasMaterial` and `HasProcess` mixin would be more straightforward, but it's not clear that it buys you much because the only consequence is that you provide an abstract method that must be implemented where you don't even know the return type.

If necessary, I could bake that chunk of framework out.

bfolie: If we can't write this method without having a holistic understanding of the GEMD object structure, then what's the value in having this logic live in `BaseObject` as opposed to citrine-python?

bfolie: Especially since `all_dependencies` is only used for batching in the context of upload, which is a citrine-python concern

gemd/entity/object/base_object.py
kroenlein: @bfolie For clarity, I did not change `|=` to `result.update()` in my last push.  Should I have?

gemd/entity/object/base_object.py
kroenlein: The mechanism has been reworked.  Does this make more sense?  I generally don't think that logic that relies only on gemd-python concepts should be implemented in a downstream package, and it seemed like implementing it as a `gemd.util` method was going to be messy.  I still would have liked a cleaner way of accessing the PropertyAndCondition elements.

gemd/entity/object/base_object.py
bfolie: No need to change

gemd/entity/object/base_object.py
bfolie: Yes, I think that this makes more sense.

bfolie (approved):

PR #156, opened 2021-11-11
Performance improvements

kroenlein:
This PR has two major changes in response to performance hits taken in our end-to-end test, plus some minor improvements:
1. The DictSerializable comparison method was rewritten to be more efficient.  This included a method to override for objects for which fields of interest aren't exactly the key-value pairs in the serialized version.
2. A substitute in place method was written because copying overhead showed up as significant in profiling.
3. Caching was also added to a few introspection methods that also showed up in the profiles.
4. The existing substitute method was discovered to be vulnerable to infinite recursion when the result of the sub() function yielded a True from applies().

These changes yielded a speedup of about 2x.

This also resolves [PLA-6423](https://citrine.atlassian.net/browse/PLA-6423)

### DIFF
--- a/gemd/entity/base_entity.py
+++ b/gemd/entity/base_entity.py
@@ -1,5 +1,5 @@
 """Base class for all entities."""
-from typing import Optional
+from typing import Optional, Dict, FrozenSet
 from collections.abc import Collection
 
 from gemd.entity.dict_serializable import DictSerializable
@@ -107,8 +107,19 @@ class BaseEntity(DictSerializable):
         return LinkByUID(scope=scope, id=uid)
 
     @staticmethod
-    def _cached_equals(this: 'BaseEntity', that: 'BaseEntity', cache: dict) -> Optional[bool]:
-        """Compute and stash whether two Base Entities are equal in a recursive sense."""
+    def _cached_equals(this: 'BaseEntity',
+                       that: 'BaseEntity',
+                       *,
+                       cache: Dict[FrozenSet, Optional[bool]] = None) -> Optional[bool]:
+        """
+        Compute and stash whether two Base Entities are equal in a recursive sense.
+
+        The cache uses ternary logic to communicate state.  True or False indicate a completed
+        evaluation.  If the cache contains None, this indicates that we have not yet completed
+        an evaluation in an earlier frame of the stack.
+        """
+        if cache is None:
+            cache = {}
         cache_key = frozenset((id(this), id(that)))
         if cache_key in cache:
             return cache[cache_key]
@@ -132,7 +143,7 @@ class BaseEntity(DictSerializable):
             this_value = this_dict[key]
             that_value = that_dict[key]
             if isinstance(this_value, BaseEntity) and isinstance(that_value, BaseEntity):
-                if BaseEntity._cached_equals(this_value, that_value, cache) is False:
+                if BaseEntity._cached_equals(this_value, that_value, cache=cache) is False:
                     cache[cache_key] = False  # Mark as failed
                     return False
             elif isinstance(this_value, Collection) and isinstance(that_value, Collection) \
@@ -158,7 +169,7 @@ class BaseEntity(DictSerializable):
                     if isinstance(x, BaseEntity):
                         for i_found, y in enumerate(that_list):
                             # Unless something really broke, y is a BaseEntity
-                            result = BaseEntity._cached_equals(x, y, cache)
+                            result = BaseEntity._cached_equals(x, y, cache=cache)
                             if result is True:
                                 found = True
                                 break
@@ -190,7 +201,7 @@ class BaseEntity(DictSerializable):
             return len(other) == 2 and other[0] in self.uids and self.uids[other[0]] == other[1]
         elif isinstance(other, BaseEntity):
             # We have to be a little clever for efficiency and to avoid infinite recursion
-            return BaseEntity._cached_equals(self, other, dict())
+            return BaseEntity._cached_equals(self, other)
         else:
             return super().__eq__(other)
 
--- a/gemd/entity/base_entity.py
+++ b/gemd/entity/base_entity.py
@@ -1,5 +1,6 @@
 """Base class for all entities."""
 from typing import Optional
+from collections.abc import Collection
 
 from gemd.entity.dict_serializable import DictSerializable
 from gemd.entity.case_insensitive_dict import CaseInsensitiveDict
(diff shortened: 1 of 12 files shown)


gemd/util/impl.py
```diff
@@ -37,41 +39,50 @@ def func(base_obj):
...
+    return _cached_issubclass(obj_type, class_or_tuple)
+
 
-    def func(obj):
-        obj_type = type(obj)
-        if obj_type not in cache:
-            cache[obj_type] = isinstance(obj, class_or_tuple)
-        return cache[obj_type]
+@functools.lru_cache(maxsize=None)
+def _cached_issubclass(
```
sfriedowitz, line 66: Is this actually a slow point in the code? What kind of comparisons are the built in `issubclass` and `isinstance` doing that would make caching this so impactful?

gemd/util/impl.py
```diff
@@ -98,34 +109,118 @@ def _substitute(thing: Any,
...
     if thing.__hash__ is not None:
         visited[thing] = new
-    if new.__hash__ is not None:
-        visited[new] = new
 
-    # assert type(thing) == type(new), "{} is not {}".format(type(thing), type(new))
     return new
 
 
+def _substitute_inplace(thing: Any,
```
sfriedowitz, line 140: Do we have support for deep-copying GEMD objects?

My naive way to implement this would be to have `_substitute_inplace`, and then have `_substitute` call a copy of the passed object, and then in-place substitute on the copy.

In terms of code complexity, this reduces the problem to a single implementation. But that may defeat the performance aspects in some way if copying is too expensive.

gemd/util/impl.py
```diff
@@ -98,34 +109,118 @@ def _substitute(thing: Any,
...
+                        sub: Callable[[object], object],
+                        applies: Callable[[object], bool],
+                        visited: Dict[object, object] = None) -> object:
```
sfriedowitz, line 143: I guess this should be `visited: Optional[Dict[object, object]]`

gemd/util/impl.py
```diff
@@ -98,34 +109,118 @@ def _substitute(thing: Any,
...
+    elif _cached_isinstance(thing, DictSerializable):
+        for k, v in thing.as_dict().items():  # Assume key can't change b/c it's an attribute
+            new_v = _substitute_inplace(v, sub, applies, visited)
+            if id(v) != id(new_v):
+                _setter_by_name(type(thing), k)(thing, new_v)
+
+    return thing
+
+
+@functools.lru_cache(maxsize=None)
```
sfriedowitz, line 205: This applies here and elsewhere in the PR.

I think that `@functools.lru_cache(maxsize=None)` is equivalent to just `@functools.cache`. Should we just switch to that in all places?

gemd/util/impl.py
```diff
@@ -98,34 +109,118 @@ def _substitute(thing: Any,
...
+def _setter_by_name(clazz: type, name: str) -> Callable:
+    """Internal method to get the setter method for an attribute."""
+    from gemd.entity.object import IngredientRun
+
+    def _emulator(inner_name: str) -> Callable:
+        return lambda self, value: setattr(self, inner_name, value)
+
+    prop = next((x[1] for x in getmembers(clazz) if x[0] == name), None)
```
sfriedowitz, line 213: I'm trying to parse what this method does and am having a really tough time doing so.

It seems that since `IngredientRun` has a separate _name method, you need to generate a setter for calls in the above methods?

Couldn't we just change `IngredientRun` to conform and use some syntactic sugar to be non-breaking? I think this may be too clever for where there seems like an easier fix by just changing IngredientRun.

sfriedowitz (commented):
The tests given me confidence that the new `_cached_equals` is correct. However, I have a handful of questions regarding the implementation in some places that should be clarified.

gemd/entity/base_entity.py
```diff
@@ -105,14 +106,93 @@ def to_link(self,
 
         return LinkByUID(scope=scope, id=uid)
 
+    @staticmethod
```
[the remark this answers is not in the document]
kroenlein: Because it uses ternary logic.  If there's another instance of the object being evaluated earlier in the stack (e.g., `process.ingredients[0].process`), it returns None to indicate unknown status.  That output changes once you've fully traversed an object.  I should document this behavior.

gemd/entity/object/material_run.py
```diff
@@ -44,7 +44,7 @@ class MaterialRun(BaseObject):
 
     typ = "material_run"
 
-    skip = {"_measurements", "_active_comparison"}
+    skip = {"_measurements"}
```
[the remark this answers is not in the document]
kroenlein: Yes, though it has consequences elsewhere because serialization is also used for tasks like object duplication.

gemd/util/impl.py
kroenlein: Just an absurdly high number of them.  This was highlighted as a slow point in previous profiling.  For this PR, I'm just swapping to a less hand-rolled version for legibility.

As an experiment, I encourage you to run the test suite with and without the caching decorator.

gemd/util/impl.py
kroenlein: I agree with the sentiment about having a large number of methods that perform similar tasks, but the deep copy methods go through JSON serialization, which is a remarkably slow task.  There's a lot of space for housecleaning in this domain across gemd-python & citrine-python, but I don't think that refactor is quite justified yet.

gemd/util/impl.py
kroenlein: That was introduced in 3.9.
https://docs.python.org/3.10/library/functools.html#functools.cache

gemd/util/impl.py
kroenlein: Ooof.

`BaseEntity` has a name property with a setter and getter.  Since you are not supposed to set `IngredientRun.name` directly, the method is overridden with a method that raises an `AttributeError`.  This is further complicated by the fact that the object will contain a `_name` value when being retrieved from the server or after a link to an `IngredientSpec` is broken, since in those cases it has a name, it just can't be accessed dynamically.

We might be able to achieve the situation described by editing the `property` in the class members, but I recall having difficult implementing that when this first raised its head.  If we could get past that, this could be handled with:
```
    elif prop.fset is None:
        setter = _emulator(f"_name")
```

gemd/entity/dict_serializable.py
```diff
@@ -135,10 +143,14 @@ def _name_repr(self, entity):
             name = getattr(entity, 'name', '<unknown name>')
             return "<{} '{}'>".format(type(entity).__name__, name)
 
+    def _dict_for_compare(self):
```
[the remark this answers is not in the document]
kroenlein: I read `fields` as a list, not as the key-value pairs we want to compare.

gemd/util/impl.py
kroenlein: Addendum: I just quickly looked into ^^.  Because the `__init__` method of `BaseAttribute` calls the name setter.  `name` is also a required argument of `BaseAttribute`.  It's functionally resolvable if we modify that to 
```
        if not isinstance(self, IngredientRun):
            self.name = name
```
but a parent class inheriting from a child is pretty hideous as well.

gemd/util/impl.py
kroenlein: It's a weird case because the user should never provide the `visited` argument, and None is an internal signifier that this is the first frame.

gemd/entity/base_entity.py
```diff
@@ -105,14 +106,104 @@ def to_link(self,
...
+                this_list = list(this_value)
+                that_list = list(that_value)
+                if len(this_list) != len(that_list):
+                    # If an object was flattened, certain lists may be empty
+                    if len(this_list) == 0:
+                        if len(this_uids) > 0:  # But if it was flattened, uids will save us
+                            continue
+                    if len(that_list) == 0:
+                        if len(that_uids) > 0:  # But if it was flattened, uids will save us
+                            continue
```
bfolie, line 161: I don't understand the scenario here.

gemd/entity/base_entity.py
```diff
@@ -105,14 +106,104 @@ def to_link(self,
...
+                            # Unless something really broke, y is a BaseEntity
+                            result = BaseEntity._cached_equals(x, y, cache=cache)
+                            if result is True:
+                                found = True
+                                break
+                            elif result is None:
+                                # Don't know yet; pass as False will appear elsewhere
+                                found = None
+                    else:
+                        found = x in that_list
```
bfolie, line 180: why do we not set `i_found` in this branch?

gemd/entity/tests/test_entity.py
```diff
@@ -29,3 +29,43 @@ def test_to_link():
...
+
+    two = ProcessSpec("Object", tags=["tags!"], uids={"scope": "id"}, notes="Notes!")
+    assert one != two, "Objects don't match unless the fields do"
+    one.notes = "Notes!"
+    assert one == two, "And then they will"
+
+    # It ignores the ingredient length mismatch if the uids matched
+    one_ing = IngredientSpec("Ingredient", process=one)
+    assert one == two
+    assert two == one
```
bfolie, line 57: Why do we want this behavior?

gemd/util/tests/test_substitute_objects.py
```diff
@@ -53,3 +54,120 @@ def func(base_ent):
...
+    rebuild = substitute_objects(links, index, inplace=True)
+    rebuilt_root = next(x for x in rebuild if x.name == root.name and x.typ == root.typ)
+    all_objs = recursive_flatmap(rebuilt_root,
+                                 func=lambda x: [x],
+                                 unidirectional=False
+                                 )
+    unique = [x for i, x in enumerate(all_objs) if i == all_objs.index(x)]
+    assert not any(isinstance(x, LinkByUID) for x in unique), "All are objects"
+    for x in unique:
+        print(x.typ, x.name)
```
bfolie, line 110: leftover debugging statement?

bfolie (commented):
I have a question about the applicability of `_cached_equals`. It's only called by `__eq__` and always called initially with an empty cache. This means that the cache is built up as we traverse a single material history, but there's no cumulative benefit if we are doing lots of equality checks.

Am I imagining an unrealistic use case, or would persisting the cache be useful in some scenarios?

gemd/entity/base_entity.py
kroenlein: For example, when you call `flatten`, the MaterialRuns in the resulting list will have a depopulated measurements list.  It is still referring to the same object, so we're ignoring that discrepancy in the comparison.  In order to have gotten to that state, however, at least one uid would have been assigned, so we can rely on uids matching if one list is empty.

gemd/entity/base_entity.py
kroenlein: Funnily enough, that's how I wrote it originally.  And then python explained that `in` is an implicit equality comparison, resulting in infinite descent.

gemd/entity/tests/test_entity.py
kroenlein: Do you mean why am I testing transitivity?  Code execution pathways vary for the left & right arguments.

This isn't technically the behavior we want to test, but getting to that state via the correct path seemed more fiddly than was needed.

kroenlein:
> I have a question about the applicability of `_cached_equals`. It's only called by `__eq__` and always called initially with an empty cache. This means that the cache is built up as we traverse a single material history, but there's no cumulative benefit if we are doing lots of equality checks.
> 
> Am I imagining an unrealistic use case, or would persisting the cache be useful in some scenarios?

The problem is that these are mutable objects, so any time you leave the immediate context an object could have mutated.  In order to implement an actual cache, `__hash__` would need to be sensitive changing contents, which in turn would make  `__hash__` much more expensive to compute.

So I think it'd be nice, but I don't think it'd end up a net win.

gemd/entity/tests/test_entity.py
bfolie: I was referring to the "ingredient length mismatch" scenario. This is essentially the same question I asked above, so we can discuss is there.

bfolie (approved):

gemd/util/impl.py
sfriedowitz: This particular block of code  is what I'm hung up on. I don't want to block this PR on it, but I would urge that this be followed up on/ticketed as a piece of maintenance work. It seems like that minor issue is necessitating pretty confusing/hacky workarounds in a case like this, and that will continue to be a hindrance if not sorted out.

sfriedowitz (approved):
Looks good, with a non-blocking comment to take note of.

PR #157, opened 2021-11-16
Cleaned up inheritence on setters

kroenlein:
This moves a unit specific check out of the `_substitute_inplace` method and adds a conditional to the `BaseObject` constructor to identify whether to set the name.

### DIFF
--- a/gemd/util/impl.py
+++ b/gemd/util/impl.py
@@ -1,7 +1,6 @@
 """Utility functions."""
 import uuid
 import functools
-from inspect import getmembers
 from typing import Dict, Callable, Union, Type, Tuple, List, Any, Optional
 from warnings import warn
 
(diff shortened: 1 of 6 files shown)


gemd/util/impl.py
```diff
@@ -197,26 +201,41 @@ def _key(obj):
...
+                _setter_by_attribute(type(thing), k)(thing, new_v)
 
     return thing
 
 
 @functools.lru_cache(maxsize=None)
-def _setter_by_name(clazz: type, name: str) -> Callable:
-    """Internal method to get the setter method for an attribute."""
-    from gemd.entity.object import IngredientRun
+def _setter_by_attribute(clazz: type, attribute: str) -> Callable:
```
sfriedowitz, line 210: Much more clear as to what's happening here now :D

sfriedowitz (approved):
LGTM

PR #159, opened 2022-01-03
Simplify import interface

kroenlein:
This PR adds the ability to import `gemd.entity` members from a variety of package levels, including supporting `from gemd import *` to get all entities that a user would normally interact with.

This PR also defines `__all__` lists in the `__init__.py` files to keep the namespace from getting too polluted from `import *`.  This is especially important because the `property.py` file was clobbering the Python `@property` built-in, causing some very confusing bugs.

### DIFF
--- a/gemd/__init__.py
+++ b/gemd/__init__.py
@@ -1,1 +1,30 @@
 """Data concepts library."""
+# flake8: noqa
+from .entity import Condition, Parameter, Property, PropertyAndConditions, \
+    CategoricalBounds, CompositionBounds, IntegerBounds, \
+    MolecularStructureBounds, RealBounds, \
+    MaterialRun, MeasurementRun, ProcessRun, IngredientRun, \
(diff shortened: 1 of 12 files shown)


sfriedowitz:
This one seems pretty straight forward, and I always have issues remembering the package structure for imports, so I like this.

One question: How was the `property.py` file in GEMD colliding with `@property` ? Was it because people were explicitly importing `import gemd.property` and qualifying objects with the `property.Object` namespace?

kroenlein:
> This one seems pretty straight forward, and I always have issues remembering the package structure for imports, so I like this.
> 
> One question: How was the `property.py` file in GEMD colliding with `@property` ? Was it because people were explicitly importing `import gemd.property` and qualifying objects with the `property.Object` namespace?

When users said `from gemd.entity.attribute import *` that imported the `property` namespace.  There were a few other related pitfalls.

sfriedowitz (approved):
LGTM

PR #160, opened 2022-01-03
Add ingest helper routines

kroenlein:
This PR adds some helper routines for creating GEMD objects according to common creation patterns, simplifying some of the calls and reducing the clerical work associated with linking the right things.

It also moves the contents of the old `ingest` directory (some very simple demo scripts) into the `demo` directory, and adds a README to the `demo` directory providing some basic information about what all demo scripts are doing.


gemd/ingest/impl.py
```diff
@@ -0,0 +1,208 @@
...
+    Returns
+    --------
+    MaterialRun
+        A MaterialRun with linked processes, specs and templates
+
+    """
+    if name is None:
+        name = input_material.name
+    my_ingredient_spec = IngredientSpec(name=name,
+                                        process=output_material.spec.process,
```
bfolie, line 101: Do we need to check that `output_material.spec` is not `None`?

gemd/ingest/impl.py
```diff
@@ -0,0 +1,208 @@
...
+    output_material: MaterialRun
+        The `process` for the returned IngredientRun will be
+        `output_material.process`
+    name: str
+        The ingredient name.  Defaults to `input_material.name`.
+
+    Returns
+    --------
+    MaterialRun
+        A MaterialRun with linked processes, specs and templates
```
bfolie, line 95: It looks like it returns the ingredient spec

gemd/ingest/impl.py
```diff
@@ -0,0 +1,208 @@
...
+                                        )
+    my_ingredient_spec = IngredientRun(spec=my_ingredient_spec,
+                                       process=output_material.process,
+                                       material=input_material
+                                       )
+    return my_ingredient_spec
+
+
+def add_measurement(material: MaterialRun,
+                    name: str = None,
```
bfolie, line 112: Shouldn't this argument come after the asterisk?

gemd/ingest/impl.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    if name is None:
+        if template is None:
+            raise ValueError("Either a name or a template must be provided")
+        else:
+            name = template.name
+    my_measurement_spec = MeasurementSpec(name, template=template)
+    my_measurement_run = MeasurementRun(name, spec=my_measurement_spec, material=material)
+
+    return my_measurement_run
```
bfolie, line 143: How does the user assert the value of the measurement? I suppose it can be added later, but that seems like an important enough part of adding the measurement to do it inside this helper method.

gemd/ingest/impl.py
```diff
@@ -0,0 +1,208 @@
...
+    attribute = attr_class(name=template.name, value=value, template=template)
+
+    if isinstance(target, MaterialSpec):
+        if attr_class is Property:
+            target.properties.append(PropertyAndConditions(property=attribute))
+        elif attr_class is Condition:
+            if len(target.properties) == 0:
+                raise ValueError("Cannot add a condition to a MaterialSpec "
+                                 "before it has at least one property.")
+            target.properties[-1].conditions.append(attribute)
```
bfolie, line 195: I don't like the implicit logic here (you must add the property and then add each conditions, then add the second property and then add each of its conditions, etc.). But I don't see a clearly better way to do it, given the nature of `PropertyAndConditions`.

bfolie:
Does it make sense to have these files in the `ingest` directory? It's true that ingest is the ultimate goal, but from the point of view of this repo they're more like builders.

gemd/ingest/impl.py
kroenlein: Maybe?  That would imply that they've gone outside the ingest framework to generate the Material Run, or are modifying something that's been flattened.  Probably enough just to throw an Exception we haven't received a full node?

gemd/ingest/impl.py
kroenlein: It does.  Good catch on the copy-pasta.

gemd/ingest/impl.py
kroenlein: I was thinking it made sense to support the optional second argument -- the most common use case felt like `add_measurement(node, 'Density')` -- but if that doesn't feed like the correct flow to you, moving the argument is reasonable.  It's always easier to add a positional argument than remove one.

gemd/ingest/impl.py
kroenlein: I would seem natural to let a user hand in a property at the same time as creating a measurement (having a measurement w/o any properties would be weird).  Since this would be an implicit `add_attribute`, you'd like to see the addition of a value and a property_template argument?

kroenlein:
> Does it make sense to have these files in the `ingest` directory? It's true that ingest is the ultimate goal, but from the point of view of this repo they're more like builders.

I could see calling the directory `builders` instead.  I certainly think the files that used to be in the ingest directory definitely belong in the same directory as the demo files.

gemd/ingest/impl.py
bfolie: > That would imply that they've gone outside the ingest framework to generate the Material Run, or are modifying something that's been flattened

Both are plausible. I could this framework being part of a hybrid workflow.

> Probably enough just to throw an Exception

Agreed. We just want to give a message like "material run needs a spec" instead of "None does not have .process attribute"

gemd/ingest/impl.py
bfolie: It's not obvious to me whether it would be more common to specify the measurement name via a string or via a template.

gemd/ingest/impl.py
bfolie: > Since this would be an implicit add_attribute, you'd like to see the addition of a value and a property_template argument?

The user could instead provide an `Attribute` (or a list of `Attributes`) directly.

Providing a `value` and `property_template` means that it makes use of the existing logic in `add_attribute`. There's less code duplication and the user doesn't have to construct the `Attribute`, which can be nice if they want to use a primitive. The downside is that the user can only assign a single value to the measurement. Additional values would have to go through `add_attribute`, which is a clunky workflow. This would only be really valuable if most measurements have a single `Attribute`.

Providing a list of `Attributes` means that the measurement can be constructed in a single call, but the user has to make all of the `Attribute`s. There would also be some additional logic to assign each `Attribute` to the appropriate collection.

Doing nothing means that the workflow is more verbose but it's consistent. You always add measurements and then add attributes.

gemd/ingest/impl.py
kroenlein: Done

gemd/ingest/impl.py
kroenlein: Changed per request.

gemd/ingest/impl.py
kroenlein: I added a `make_attribute` method and an `attributes` argument to add_measurement to scratch this itch and try to still keep methods simple and consistent.  Let me know if the change makes sense.

gemd/builders/impl.py
```diff
@@ -0,0 +1,424 @@
...
+                my_measurement_run.conditions.append(attribute)
+            elif isinstance(attribute, Parameter):
+                my_measurement_run.parameters.append(attribute)
+            else:
+                raise ValueError(f"Unhandled Attribute type {type(attribute)}")
+
+    return my_measurement_run
+
+
+@overload
```
bfolie, line 244: The heavy use of `@overload` seems unnecessary. To take `add_attribute` as an example, the multiple overloaded methods enforce "A `PropertyTemplate` goes with a `HasProperties`, ..." but the implementation still has to check that the template is right for the target, so I don't see what is gained.

kroenlein: Yeah, I was trying out a new toy.  I agree that it didn't return the quality of results I was hoping for.  At least for IntelliJ, the integration with the automated type checker didn't do quite as much I was hoping it would.  The better way to do this might be Generics, but even that probably doesn't buy us enough to justify the all the code.

bfolie (approved):

PR #161, opened 2022-01-03
Add local validation framework for GEMD objects

kroenlein:
Right now, all validation is performed server side with no constraints or feedback to users client side when types or values are inconsistent with templates.  This PR does two things:

1. It adds automatic validation code at every level of value assignment so that a user can receive feedback immediately when a bound is violated.
2. It adds the ability for the user to control whether the inconsistencies with bounds are ignored, are a warning, or are fatal.  The default here is Warning, which will make existing scripts noisier but will not break existing workflows.  It is proposed that we consider changing the default to Fatal when we release gemd-python 2.0.

Note that there is no guarantee that an object can be validated since any template could just be referenced by LinkByUID,  in which case we have no local knowledge of its bounds.

### DIFF
--- a/gemd/entity/bounds_validation.py
+++ b/gemd/entity/bounds_validation.py
@@ -32,7 +32,7 @@ def set_validation_level(level: WarningLevel):
 
 
 @contextmanager
-def validation_context(level: WarningLevel):
+def validation_level(level: WarningLevel):
     """Provide a context for setting a WarningLevel locally."""
     global BOUNDS_VALIDATION
     # Swap values and store
--- a/gemd/entity/object/has_parameters.py
+++ b/gemd/entity/object/has_parameters.py
@@ -1,16 +1,14 @@
 """For entities that have parameters."""
 from gemd.entity.has_dependencies import HasDependencies
+from gemd.entity.object.has_template_check_generator import HasTemplateCheckGenerator
 from gemd.entity.template.has_parameter_templates import HasParameterTemplates
 from gemd.entity.attribute.parameter import Parameter
 from gemd.entity.setters import validate_list
 
-from abc import abstractmethod
-from typing import Union, Iterable, List, Set, Callable, TypeVar
-
-T = TypeVar('T')
+from typing import Union, Iterable, List, Set
 
 
-class HasParameters(HasDependencies):
+class HasParameters(HasTemplateCheckGenerator, HasDependencies):
     """Mixin-trait for entities that include parameters.
 
     Parameters
@@ -28,10 +26,6 @@ class HasParameters(HasDependencies):
     def parameters(self) -> List[Parameter]:
         """Get the list of parameters."""
         return self._parameters
-
-    @abstractmethod
-    def _generate_template_check(self, validate: Callable[[T], bool]) -> Callable[[T], T]:
-        """Generate a closure for the object and the validation routine."""
 
     @parameters.setter
     def parameters(self, parameters: Iterable[Parameter]):
--- a/gemd/entity/object/has_parameters.py
+++ b/gemd/entity/object/has_parameters.py
@@ -1,4 +1,5 @@
 """For entities that have parameters."""
+from gemd.entity.template.has_parameter_templates import HasParameterTemplates
 from gemd.entity.attribute.parameter import Parameter
 from gemd.entity.setters import validate_list
 from gemd.entity.bounds_validation import get_validation_level, WarningLevel
@@ -30,7 +31,8 @@ class HasParameters(object):
             # if Has_Templates hasn't been called yet, it won't have a _template attribute
             template = getattr(self, "template", None)
             level = get_validation_level()
-            accept = level == WarningLevel.IGNORE or template is None \
+            accept = level == WarningLevel.IGNORE \
+                or not isinstance(template, HasParameterTemplates) \
                 or self.template.validate_parameter(x)
 
             if not accept:
--- a/gemd/entity/object/has_parameters.py
+++ b/gemd/entity/object/has_parameters.py
@@ -3,10 +3,11 @@ from gemd.entity.has_dependencies import HasDependencies
 from gemd.entity.template.has_parameter_templates import HasParameterTemplates
 from gemd.entity.attribute.parameter import Parameter
 from gemd.entity.setters import validate_list
-from gemd.entity.bounds_validation import get_validation_level, WarningLevel
-from gemd.entity.dict_serializable import logger
 
-from typing import Union, Iterable, List, Set
+from abc import abstractmethod
+from typing import Union, Iterable, List, Set, Callable, TypeVar
+
+T = TypeVar('T')
 
 
 class HasParameters(HasDependencies):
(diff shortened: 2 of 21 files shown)


sfriedowitz:
I am going to start looking at this one now @kroenlein

bfolie:
From a quick glance, the logic looks fine. I think that the relevant discussion here is not how to do local validation but whether to do it at all.

There are good reasons to only do validation server-side: it localizes the validation logic in one place and the server has access to all of the relevant information, while the client may only know about link-by-uids. So if we're going to add client-side validation code then there should be a clear and compelling reason. What's the relevant user story for which this is a significant benefit? Is there a JIRA ticket? And if there is a relevant user story, could be solve the problem with a faster failure and/or more clear error message from the server?

kroenlein:
> From a quick glance, the logic looks fine. I think that the relevant discussion here is not how to do local validation but whether to do it at all.
> 
> There are good reasons to only do validation server-side: it localizes the validation logic in one place and the server has access to all of the relevant information, while the client may only know about link-by-uids. So if we're going to add client-side validation code then there should be a clear and compelling reason. What's the relevant user story for which this is a significant benefit? Is there a JIRA ticket? And if there is a relevant user story, could be solve the problem with a faster failure and/or more clear error message from the server?

https://citrine.atlassian.net/browse/PLA-5806

gemd/entity/bounds_validation.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def set_validation_level(level: WarningLevel):
+    """Set the value of the BOUNDS_VALIDATION."""
+    global BOUNDS_VALIDATION
+    BOUNDS_VALIDATION = WarningLevel(level)
+
+
+@contextmanager
+def validation_context(level: WarningLevel):
```
sfriedowitz, line 35: Defining custom contexts with this decorator is something I haven't used before, so I'm mainly just curious.

Is this used in the form of `with validation_context(level): do things...` ?

If so, would naming it `validation_level` fit better gramatically?

gemd/entity/object/has_parameters.py
```diff
@@ -24,4 +27,20 @@ def parameters(self):
 
     @parameters.setter
     def parameters(self, parameters):
-        self._parameters = validate_list(parameters, Parameter)
+        def _template_check(x: Parameter) -> Parameter:
```
sfriedowitz, line 30: This internal method `_template_check` is roughly identical to the above for conditions. Can we abstract it out and pass the `accept` condition as a lambda function maybe?

gemd/entity/bounds_validation.py
kroenlein: You are correct in how it is used, and your suggestion makes sense grammatically.  My only hesitation would be if having `get_validation_level`, `set_validation_level` and `validation_level` methods would be confusing.  I'm inclined to say `It'll be fine.`

gemd/entity/object/material_spec.py
```diff
@@ -58,7 +61,25 @@ def properties(self):
 
     @properties.setter
     def properties(self, properties):
-        self._properties = validate_list(properties, PropertyAndConditions)
+        def _template_check(x: PropertyAndConditions) -> PropertyAndConditions:
```
sfriedowitz, line 64: Same thing. This is identical to `conditions` and `parameters`, and could probably be re-used.

bfolie:
> Sorry, I should have included this in the initial description: https://citrine.atlassian.net/browse/PLA-5806

One of the last comments on that ticket is "This card was iced because this is best accomplished in the backend ("client validation is no validation")". Is this actually best accomplished in the client or are you doing it because nobody is picking up the backend work?

sfriedowitz (commented):
I didn't study the tests in excruciating detail, but the validation logic seems okay with me. 

I think it would be worth while to find a way to reduce code duplication on those internal methods. I also think `validation_level` makes more sense as the context manager if you feel so inclined to make that change.

kroenlein:
> > Sorry, I should have included this in the initial description: https://citrine.atlassian.net/browse/PLA-5806
> 
> One of the last comments on that ticket is "This card was iced because this is best accomplished in the backend ("client validation is no validation")". Is this actually best accomplished in the client or are you doing it because nobody is picking up the backend work?

We now have register_all w/ a validation check (dry_run) and most of its behaviors around large checks have been resolved with the batching work.  This is better than what existed at the initial time of request.  In order to do better, we would need to increase the batch size to arbitrarily large (10k?).  So you are still left with a situation where an ingestor must run for hours before you get feedback that your cell was misformatted (a test that could be done in 10 minutes for a large dataset).  As well, the information returned by the platform is well-separated from the actual data error, making it much harder to understand what actually needs fixing.

tldr; I'd say it's better accomplished in the client, and we've pushed server side to as good as it can get.

gemd/entity/object/has_parameters.py
kroenlein: The methods are very similar, but would require a closure for construction since (ignoring the type checking) they depend on `HasXXXXTemplates` and `validate_XXXX`.  There's also no mixin parent class here, though I can certainly build one (or just create a local util package) if DRY suggests this is the wiser route.

kroenlein: I've created a centralized _template_check closure generator in HasTemplates, and then has HasSpec piggy back on it since I didn't want to implement it twice.

gemd/entity/object/material_spec.py
kroenlein: Addressed as well.

gemd/entity/object/has_template_check_generator.py
```diff
@@ -0,0 +1,71 @@
...
+        function(Attribute)
+            If the attribute is inconsistent, ignores, warns or dies based on get_validation_level
+
+        Raises
+        ------
+        ValueError
+            If `value` is not one of the allowed types or if mapping types fails.
+
+        """
+        # Determine which class `validate` is from, so we can type check the object template
```
sfriedowitz, line 44: I think there is a comment missing at the top of this block here that explains the problem that this module inspection stuff is solving.

As far as I understand it, you are passing a `validate` function which is referencing a method to some class. You need to check which module you are in and find the class type that is passing the function and use that class type in the closure you are generating. Is that correct? If so, at the top of this blob I think you should write clearly the problem this is solving. 

The inline comments help, but clearly stating the problem that this is addressing will make it easier to read for the next maintainer.

gemd/entity/object/has_template_check_generator.py
```diff
@@ -0,0 +1,71 @@
...
+
+class HasTemplateCheckGenerator(ABC):
+    """Mix-in trait for objects that can generate a Check Template closure."""
+
+    @property
+    @abstractmethod
+    def template(self) -> Union[BaseTemplate, LinkByUID]:
+        """Get the object template associated with this object."""
+
+    def _generate_template_check(self,
```
sfriedowitz, line 22: Can you add in this docstring a comment describing the signature of the returned closure function please? The comment `Generate a closure for the object and the validation routine.` is not as clear as one would hope.

E.g. "The generated closure references the class type that calls this method, and takes as argument a value to validate. The value is checked against blah blah..."

sfriedowitz (changes requested):
I think the inheritance structure is much improved with this iteration.

I am requesting some additional comments on the critical method so that the next reader has an easier time with this.

gemd/entity/object/has_template_check_generator.py
```diff
@@ -0,0 +1,89 @@
...
+                                 validate: Callable[["HasTemplateCheckGenerator", T], bool]
+                                 ) -> Callable[[T], None]:
+        """
+        Generate a closure for the object and the validation routine.
```
sfriedowitz, line 26: 🙇

sfriedowitz (approved):
LGTM

PR #162, opened 2022-01-06
Update Travis to test up to 3.10

kroenlein:
Python 3.10 is currently considered stable, so we should be testing against it.  This augments the tests that Travis runs to do this, and propagates the need to support Python 3.10 to required packages.


kroenlein:
Almost 12 minutes is required for 3.11-dev to build pandas, which is required for 2 tangential tests.  This may motivate dropping 3.11-dev from the proposed suite of tests.

bfolie:
> Almost 12 minutes is required for 3.11-dev to build pandas, which is required for 2 tangential tests. This may motivate dropping 3.11-dev from the proposed suite of tests.

I agree. It's highly unlikely that our customers are using the in-development version of python, and we can cross any relevant bridges when it is officially released.

.travis.yml
```diff
@@ -1,8 +1,16 @@
 language: python
+dist: bionic
 python:
 - '3.6'
 - '3.7'
 - '3.8'
+- '3.9'
+- '3.10.0'
+- '3.10.1'
```
bfolie, line 9: Why explicitly test both 3.10.0 and 3.10.1?

.travis.yml
```diff
@@ -1,8 +1,16 @@
 language: python
+dist: bionic
 python:
 - '3.6'
```
bfolie, line 4: Weren't we going to drop 3.6?

.travis.yml
kroenlein: Because there is no Travis option for just `3.10`; only `3.10.0`, `3.10.1` and `3.10-dev`.  I'm okay with dropping `3.10.0` but had it in there because conda doesn't have a 3.10.1 build yet so I wanted to be able to make sure that anything that went wrong was not because of a 3.10.0 v. 3.10.1 difference.

https://docs.travis-ci.com/user/languages/python/

.travis.yml
kroenlein: One of our customers is currently using 3.6 to run their ingest pipeline.  The customer team is following up to understand why that is, but so long as it is we need to support it.

.travis.yml
bfolie: To clarify, nothing went wrong with 3.10.1, correct?

bfolie (approved):

PR #166, opened 2022-05-04
as strings are iterables this was not working as intended

lkubie:
If a user passes a string into what should be a list, make a list where the only element is the string.


kroenlein (changes requested):
Generally looks good.  Can we get a test?

lkubie (commented):
Thanks for the test!!

kroenlein (approved):
With tests added, this looks good.

PR #168, opened 2022-05-17
Correct spelling, type-checks and dependencies

kroenlein:
A substantial number of corrections to spelling, type checks, dependencies and other sundries identified by a thorough review with support from my IDE.

There should be no functional changes in this PR.


gemd/entity/object/has_template_check_generator.py
```diff
@@ -21,7 +21,7 @@ def template(self) -> Union[BaseTemplate, LinkByUID]:
 
     def _generate_template_check(self,
                                  validate: Callable[["HasTemplateCheckGenerator", T], bool]
-                                 ) -> Callable[[T], None]:
+                                 ) -> Callable[[T], Any]:
```
bfolie, line 24: Why `Any`, especially since the doc string says it returns a function from `Attribute` to `None`?

bfolie (approved):

gemd/entity/object/has_template_check_generator.py
kroenlein: Nice catch; I got too aggressive in copying the `validate_list` signature up.

bfolie (approved):

PR #169, opened 2022-05-23
Move cached_instance to public method

kroenlein:
In anticipation of releasing citrine-python v1.31, moving a private method that citrine-python references to a public method.

https://github.com/CitrineInformatics/citrine-python/blob/568a03de0c7f45f7be6ea80f029e51fa6eff70ed/src/citrine/_serialization/properties.py#L13

No prior public release of citrine-python references this method.

### DIFF
--- a/gemd/util/tests/test_cached_isinstance.py
+++ b/gemd/util/tests/test_cached_isinstance.py
@@ -1,0 +1,30 @@
+from gemd.util import cached_isinstance
+
+from typing import List, Iterable
+
+
+def test_cached_isinstance():
+    """Verify that cached_isinstance is operating as exepcted."""
+    passing_pairs = [
+        ("one", str),
+        (1, int),
+        ("one", (str, int)),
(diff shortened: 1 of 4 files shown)


gemd/util/tests/test_cached_isinstance.py
```diff
@@ -0,0 +1,30 @@
+from gemd.util import cached_isinstance
+
+from typing import List, Iterable
+
+
+def test_cached_isinstance():
```
bfolie, line 6: Did we not already have a test like this?

bfolie, line 6: Did we not already have a test like this?

bfolie (approved):

gemd/util/tests/test_cached_isinstance.py
kroenlein: We did not because it was never publicly exposed.

PR #170, opened 2022-07-14
Add union and update methods to bounds

kroenlein:
This PR adds a method to `bounds` objects that allows computation of mergers of one or many bounds objects, or extension of a given bounds object to accept a given value.

* For ordered / range-based sets, it expands such that the new range encompasses all passed ranges.
* For unordered sets, it adds the smallest number of new elements to include all passed sets.

`union` returns a new bounds without changing the old; `update` changes the bounds in place.

### DIFF
--- a/gemd/entity/bounds/categorical_bounds.py
+++ b/gemd/entity/bounds/categorical_bounds.py
@@ -95,7 +95,8 @@ class CategoricalBounds(BaseBounds):
             misses = {type(x).__name__
                       for x in others
                       if not isinstance(x, (CategoricalBounds, CategoricalValue))}
-            raise TypeError(f"union requires consistent typing; found {misses}")
+            raise TypeError(f"union requires consistent typing; "
+                            f"expected categorical, found {misses}")
         result = self.categories.copy()
         for bounds in others:
             if isinstance(bounds, CategoricalValue):
--- a/gemd/entity/bounds/categorical_bounds.py
+++ b/gemd/entity/bounds/categorical_bounds.py
@@ -2,7 +2,7 @@
 from gemd.entity.bounds.base_bounds import BaseBounds
 from gemd.entity.util import array_like
 
-from typing import Union
+from typing import Union, Set, Optional, Iterable
 
 
 class CategoricalBounds(BaseBounds):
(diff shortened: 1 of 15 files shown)


gemd/entity/bounds/categorical_bounds.py
```diff
@@ -70,6 +70,53 @@ def contains(self, bounds: Union[BaseBounds, "BaseValue"]) -> bool:
...
+            The union of this bounds and the passed bounds
+
+        """
+        from gemd.entity.value.categorical_value import CategoricalValue
+
+        if any(not isinstance(x, (CategoricalBounds, CategoricalValue)) for x in others):
+            misses = {type(x).__name__
+                      for x in others
+                      if not isinstance(x, (CategoricalBounds, CategoricalValue))}
+            raise TypeError(f"union requires consistent typing; found {misses}")
```
bfolie, line 98: The error for this (and the other implementations) should say which type was expected. So here for example, "union requires consistent typing; expected categorical, found {misses}"

gemd/entity/bounds/categorical_bounds.py
```diff
@@ -70,6 +70,53 @@ def contains(self, bounds: Union[BaseBounds, "BaseValue"]) -> bool:
...
+        result = self.categories.copy()
```
bfolie, line 99: Since this isn't a deep copy, won't the updates made below also affect this Bounds object?

kroenlein: It shouldn't since I'm only copying the categories, not the bounds.  I think I'm verifying this with this test:
https://github.com/CitrineInformatics/gemd-python/blob/11861577b18c99eb473084ab46bd6a56d2714a38/gemd/entity/bounds/tests/test_categorical_bounds.py#L45-L47

gemd/entity/bounds/real_bounds.py
```diff
@@ -86,6 +86,62 @@ def contains(self, bounds: Union[BaseBounds, "BaseValue"]) -> bool:
...
+            misses = {type(x).__name__
+                      for x in others
+                      if not isinstance(x, (RealBounds, ContinuousValue))}
+            raise TypeError(f"union requires consistent typing; found {misses}")
+        lower = self.lower_bound
+        upper = self.upper_bound
+        unit_ = self.default_units
+        for bounds in others:
+            if isinstance(bounds, ContinuousValue):
+                this_lo, this_hi = bounds._to_bounds()._convert_bounds(unit_)
```
bfolie, line 118: `ContinuousValue` objects get converted to the default units, but if `others` contains other `RealBounds` objects, I don't see where those would be converted

bfolie (approved):

PR #172, opened 2022-07-19
Be more accepting of ways people express InChI

kroenlein:
Improved filtering around InChI strings.  I think this meets the requested functionality.

### DIFF
--- a/gemd/entity/value/inchi_value.py
+++ b/gemd/entity/value/inchi_value.py
@@ -22,14 +22,19 @@ class InChI(MolecularValue):
 
     @property
     def inchi(self) -> str:
-        """Get the formula as a string."""
+        """Get the InChI as a string."""
         return self._inchi
 
     @inchi.setter
     def inchi(self, value: str):
+        """Set the InChI, correcting for some minor variations in format."""
         if value is None:
             self._inchi = None
         elif isinstance(value, str):
+            if not value.lower().startswith('inchi'):
+                value = f"InChI=1S/{value}"
+            elif not value.startswith('InChI'):
+                value = value.replace(value[:5], 'InChI')
             self._inchi = value
         else:
             raise TypeError("InChI must be given as a string; got {}".format(type(value)))
--- a/gemd/entity/value/inchi_value.py
+++ b/gemd/entity/value/inchi_value.py
@@ -31,7 +31,9 @@ class InChI(MolecularValue):
         if value is None:
             self._inchi = None
         elif isinstance(value, str):
-            if not value.lower().startswith('inchi'):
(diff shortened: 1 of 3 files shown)


gemd/entity/value/inchi_value.py
```diff
@@ -22,14 +22,19 @@ def __init__(self, inchi=None):
...
         return self._inchi
 
     @inchi.setter
     def inchi(self, value: str):
+        """Set the InChI, correcting for some minor variations in format."""
         if value is None:
             self._inchi = None
         elif isinstance(value, str):
+            if not value.lower().startswith('inchi'):
+                value = f"InChI=1S/{value}"
```
lkubie, line 35: I've seen this with the `1s/` so I want to check for that too.  [Here's Sigma Aldrich for example](https://www.sigmaaldrich.com/US/en/product/aldrich/757136)


```suggestion
                if not value.lower().startswith("1s/") and not value.lower().startswith("1/"):
                    value = f"InChI=1S/{value}"
                else:
                    value = f"InChI={value}"
```

gemd/entity/value/tests/test_inchi.py
```diff
@@ -14,13 +14,16 @@ def test_json():
...
-    """Check that we can set the formula."""
+    """Check that we can set the inchi."""
     test_struct = "InChI=1/C8H8O3/c1-11-8-4-6(5-9)2-3-7(8)10/h2-5,10H,1H3"
     test_inchi = InChI()
     assert test_inchi.inchi is None
     test_inchi.inchi = test_struct
     assert test_inchi.inchi == test_struct
 
+    assert InChI("inchi=1/C8H8O3").inchi == "InChI=1/C8H8O3"
+    assert InChI("C8H8O3").inchi == "InChI=1S/C8H8O3"
```
lkubie, line 25:
```suggestion
    assert InChI("C8H8O3").inchi == "InChI=1S/C8H8O3"
    assert InChI("1/C8H8O3").inchi == "InChI=1/C8H8O3"
    assert InChI("1s/C8H8O3").inchi == "InChI=1s/C8H8O3"
```

lkubie (changes requested):
Added a little robustness and provided a real-life example as to why.

gemd/entity/value/inchi_value.py
kroenlein: Great catch; thanks!

gemd/entity/value/tests/test_inchi.py
kroenlein: Good add, though the standard expects a capital S.

lkubie (approved):

PR #177, opened 2022-10-25
Cap pint at less than version 0.20.0.

anoto-moniz:
The pint.unit package was dropped in 0.20.0, breaking our builds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 
 setup(name='gemd',
-      version='1.11.0',
+      version='1.11.1',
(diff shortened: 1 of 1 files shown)


pacdaemon (approved):
LGTM

kroenlein (dismissed):
You need to bump the version on line 5.  This is probably a minor because it shouldn't be breaking.

anoto-moniz:
> You need to bump the version on line 5. This is probably a minor because it shouldn't be breaking.

Good catch! Went to 1.11.1.

kroenlein (approved):
LGTM

PR #179, opened 2022-10-27
Bump ceiling on Pint version to 0.20

kroenlein:
Increase Pint version ceiling, now that we are in compliance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ python:
 env:
 - PINT_VERSION=0.13
 - PINT_VERSION=0.20
+jobs:
+  exclude:
+    - python: 3.7
+      env: PINT_VERSION=0.20
 install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,13 +5,21 @@ python:
 - '3.8'
 - '3.9'
 - '3.10'
+env:
+- PINT_VERSION=0.13
```
anoto-moniz, line 9: Do we need to define this variable twice?

kroenlein: It permutes the tests, so you run each Python version with each Pint version.  The alternative is to explicitly define the combinations we want to test.
https://docs.travis-ci.com/user/build-matrix/

anoto-moniz: Ahhhhhh. gotcha! Makes sense.

PR #180, opened 2023-01-30
Add support for scaling factors to gemd-python

kroenlein:
[Pint](https://pint.readthedocs.io) does not support scaling factors in units (e.g., [support ticket](https://github.com/hgrecco/pint/issues/793)).  In order to provide support for units encountered in the wild (e.g., `g / 2.5 cm`), this PR adds support for strictly numeric scaling factors.

It dynamically adds a dimensionless scaling unit for any number encountered (where `Number` is defined using the same tokenizer as the core Pint library).  Multiplication between a scaling factor and a unit also bind tighter than division (i.e., if that scaling factor is preceded by division and followed by a unit, it also puts that unit in the denominator).

Internal Citrine reference: https://citrine.atlassian.net/browse/PLA-6256

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
 - '3.9'
 - '3.10'
 env:
-- PINT_VERSION=0.13
+- PINT_VERSION=0.18
 - PINT_VERSION=0.20
 jobs:
   exclude:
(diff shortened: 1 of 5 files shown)


kroenlein:
pint > 0.18 requires unit names be valid identifiers.  Moved the munged version to first position in the definition statement, which unfortunately means that the pretty format isn't as pretty as it might be.

dleblanc (approved):
LGTM

PR #182, opened 2023-04-11
Add central registry for classes; add metaclass for DictSerializable

kroenlein:
This PR is part of a series of changes to move gemd-python to a more modern object model without breaking citrine-python along the way.

Major changes:
* A custom metaclass was added to `DictSerializable` to keep track of concrete derived classes from it.  The removes the need for an explicit list maintained in the JSON module.
    * The register_classes method in gemd_json is therefore deprecated, but still functional as citrine-python relies on it.
* A `typ` and a `skip` argument have been added to the `DictSerializable` metaclass constructor, so concrete classes derived from `DictSerializable` no longer necessarily rely on an explicit `typ` variable.
    * Children of classes that used to have a `typ` string defined will continue to function unchanged.
    * Children of DictSerializable that explicitly define a `typ` string and previously used register_classes should continue to function unchanged.
    * This will possibly facilitate migrating to private variables, classmethods, and/or properties in the future without breaking child classes in dependent libraries.
    * This will also permit throwing an exception when an ostensibly abstract object's constructor is invoked.  We cannot yet effect this change because some child classes in citrine-python have not defined a `typ` string.

I would have liked to move `typ` and `skip` into methods instead of class variables, but current access patterns include retrieving the values through both the object and the class.  This means:
* a `property` on the class yields a property object when access via the class
* a `property` on the metaclass yields no value via the object
* a `classmethod` would support returning the value from both the class and the object, but requires parentheses, thus breaks current access  patterns in both cases
* Python 3.11 removes wrapping a property with a classmethod 

I'm open to suggestions for patterns that would move these into accessors without breaking existing code.


anoto-moniz (approved):

anoto-moniz (approved):

bfolie (approved):
Did you locally pull this into citrine-python and confirm that tests pass?

kroenlein:
> Did you locally pull this into citrine-python and confirm that tests pass?

Yes, plus nextgen-devkit end-to-ends.

PR #183, opened 2023-04-18
Add additional tests for unexpected fields in units

kroenlein:
Field testing showed some bad behavior:
* Punctuation like `"` and `:` were being swallowed, leading to things that weren't valid units being accepted
* Strings that were just numbers being being interpreted as valid units

### DIFF
--- a/gemd/units/impl.py
+++ b/gemd/units/impl.py
@@ -1,13 +1,13 @@
 """Implementation of units."""
 import re
 
-from pint import UnitRegistry, Unit, register_unit_format
+from pint import UnitRegistry, Unit, register_unit_format, Quantity
 from pint.compat import tokenizer
 from tokenize import NAME, NUMBER, OP
 # alias the error that is thrown when units are incompatible
 # this helps to isolate the dependence on pint
 from pint.errors import DimensionalityError as IncompatibleUnitsError  # noqa Import
-from pint.errors import UndefinedUnitError
+from pint.errors import UndefinedUnitError, DefinitionSyntaxError  # noqa Import
 
 import functools
 import pkg_resources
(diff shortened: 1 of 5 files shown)


gemd/units/tests/test_parser.py
```diff
@@ -42,12 +42,17 @@ def test_parse_unexpected():
         5,
         "cp",  # Removed because of risk of collision with cP
         "chain",  # Survey units eliminated
-        "SECONDS"  # Not just case insensitivity
+        "SECONDS",  # Not just case insensitivity
+        "lb : in^3",  # Not just case insensitivity
```
bfolie, line 46: This comment seems unrelated

kroenlein: Good eyes, copypasta

gemd/units/impl.py
```diff
@@ -26,6 +26,9 @@ def _scaling_preprocessor(input_string: str) -> str:
     tight_division = False
     scales = []
 
+    if next(token for token in tokens).type == NUMBER:
+        return input_string  # The unit can't have a leading number; scaling factors are internal
```
bfolie, line 30: Why does returning the input string lead to the desired error being thrown?

Also, why is a leading number not allowed? If `g / 100 L` is then why not `100 L / g`? Or is this to cover for the cases in which the input string is _just_ a number?

gemd/units/impl.py
```diff
@@ -36,8 +39,13 @@ def _scaling_preprocessor(input_string: str) -> str:
         if not exponent and token.type == NUMBER:
             scales.append([token.string, False])
             tight_division = division
-        exponent = token.type == OP and token.string in {"^", "**"}
-        division = token.type == OP and token.string in {"/", "//"}
+        if token.type == OP:
+            if token.string not in {"+", "-", "*", "/", "//", "^", "**", "(", ")"}:
```
bfolie, line 43: This would be more readable if the set of all allowed tokens was broken out as a constant

gemd/units/impl.py
kroenlein: We exit the preprocessor without mangling the leading digit, so it's handled as it would have been handled by default.  We could probably get equivalent functionality by just swallowing the first term, but I didn't want to have to deal with possible weirdness following from that if that first term would have set some of the internal booleans.

No leading digit because normal Pint behavior says that should be a magnitude on a value, and that kinda matches.  I personally also can't think of a customary unit that has a scaling factor that has it up front like that.  It conveniently also covers the case where it's just a number.

gemd/units/impl.py
kroenlein: Fair.  I'd wanted to import it from Pint, but it's a private variable.

bfolie (approved):
I don't fully understand it, but the tests pass and they look appropriate

bfolie (approved):

PR #185, opened 2023-04-21
Include periods in acceptable punctuation in units

kroenlein:
Add `.` to acceptable punctuation in units.


dleblanc (dismissed):

kroenlein:
> People use periods in their units?

a.u.
mol.%

Formally they shouldn't, but in practice they do.

bfolie (approved):

PR #188, opened 2023-05-08
Boost testing range

kroenlein:
This PR adds Python 3.12 to the Travis testing, adds 0.21 to the tested Pint versions, and cleans up the testing install a bit.


kroenlein:
Travis does not yet support 3.12.

PR #190, opened 2023-05-22
Augment scaling factor parsing

kroenlein:
This PR fixes some parsing issues that were reported for units with scaling factors and augments the units methods to fully cover the units-service needs.  Specifically:
* It adds a `get_base_units method`, required for BE indexing
* It adds a flag the `parse_units` method to return a Unit object instead whatever it was handed
* It rewrites the scaling factor parsing logic, which had some subtle bugs

Since this is also intended for consumption by out units web service, this PR also reduces the LRU cache sizes from unbounded to large.

### DIFF
--- a/gemd/units/impl.py
+++ b/gemd/units/impl.py
@@ -3,7 +3,7 @@ import re
 
 from pint import UnitRegistry, Unit, register_unit_format, Quantity
 from pint.compat import tokenizer
-from tokenize import NAME, NUMBER, OP
+from tokenize import NAME, NUMBER, OP, Token, ERRORTOKEN
 # alias the error that is thrown when units are incompatible
 # this helps to isolate the dependence on pint
 from pint.errors import DimensionalityError as IncompatibleUnitsError  # noqa Import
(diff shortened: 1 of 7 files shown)


gemd/units/impl.py
```diff
@@ -3,70 +3,132 @@
...
+from typing import Union, List, Tuple
 
 # use the default unit registry for now
 DEFAULT_FILE = pkg_resources.resource_filename("gemd.units", "citrine_en.txt")
 _ALLOWED_OPERATORS = {".", "+", "-", "*", "/", "//", "^", "**", "(", ")"}
 
 
+def _space_after_minus_preprocessor(input_string: str) -> str:
+    """A preprocessor that protects against a pint < 0.21 bug."""
+    return re.sub(r"(?<=-)\s+(?=\d)", "", input_string)
```
kroenlein, line 23: Inconsistent behavior between 0.20 & 0.21 for `m ** - 1` (vs. `m ** -1`).  Older version threw a parsing error, so this corrects for that bug by avoiding the messy condition.

gemd/units/impl.py
```diff
@@ -3,70 +3,132 @@
...
+
+
+def _scientific_notation_preprocessor(input_string: str) -> str:
+    """Preprocessor that converts x * 10 ** y format to xEy."""
+    def _as_scientific(matchobj: re.Match) -> str:
+        return f"{matchobj.group(1) or '1'}e{matchobj.group(2)}"
+
+    number = r'\b(?:(\d+\.?\d*|\.\d+)\s*\*\s*)?10\s*(?:\*{2}|\^)\s*\+?(-?\d+\b)'
+    return re.sub(number, _as_scientific, input_string)
```
kroenlein, line 32: Solves some parsing complexity by making all scaling factors a single word (instead of `[x, *, 10, **, y]`) when possible.

gemd/units/impl.py
```diff
@@ -3,70 +3,132 @@
...
+            else:  # Break blocks for two units in a row
+                blocks.append([token])
+        elif token.type == NUMBER:
+            blocks[-1].append(token)
+        elif token.type == ERRORTOKEN:  # Keep non-legal Python symbols like °
+            blocks[-1].append(token)
+        # Drop other tokens, such as EOF
+
+        if len(operator_stack) > 0 and operator_stack[-1].string in {"**", "^"}:
+            operator_stack.pop()  # Exit context for this exponential
```
kroenlein, line 72: This breaks the complete expression into a series of terms that are separated by multiplication or division -- essentially the basic units we might want to substitute out.

gemd/units/impl.py
```diff
@@ -3,70 +3,132 @@
...
+                block_string = input_string[block[0].start[1]:block[-1].end[1]]
+                if i_name is None:
+                    unit_string = None
+                else:
+                    unit_string = input_string[block[position + 1].start[1]:block[i_name].end[1]]
+                todo.append((block_string, value, unit_string))
+        elif len(numbers) > 1:
+            raise ValueError(
+                f"Replicate scaling factor ({[n[1] for n in numbers]}) in {input_string}"
+            )
```
kroenlein, line 99: This evaluates which of the expression fragments actually contain scaling factors we want to substitute, and stashes those cases in a todo list.

gemd/units/impl.py
```diff
@@ -3,70 +3,132 @@
...
+            long = stripped
+            short = stripped
+
+        underscored = re.sub(r"[-.]", "_", long)
+        valid = f"_{underscored}"
+        if valid not in _REGISTRY:
+            # Parse subexpression to clean things up for define
+            value = f"{_REGISTRY.parse_expression(scaled_term)}"
+            _REGISTRY.define(f"{valid} = {value} = {short}")
+        input_string = re.sub(regex, valid, input_string)
```
kroenlein, line 122: This normalizes the scaling factor expressions, and if they are novel, stores them in the unit registry.

gemd/units/impl.py
```diff
@@ -75,9 +137,9 @@ def _format_clean(unit, registry, **options):
     numerator = []
     denominator = []
     for u, p in unit.items():
-        if re.match(r"_[\d_]+$", u):
-            # Munged scaling factor; drop leading underscore, restore . and -
-            u = re.sub(r"(?<=\d)_(?=\d)", ".", u[1:]).replace("_", "-")
+        if re.match(r"_[\d_]+", u):
+            # Munged scaling factor; grab symbol, which is the prettier
+            u = registry.get_symbol(u)
```
kroenlein, line 142: Use the symbol rather than trying to un-munge a number with regular expressions.

gemd/units/impl.py
```diff
@@ -98,7 +160,7 @@ def _format_clean(unit, registry, **options):
 
 
 @functools.lru_cache(maxsize=1024)
-def parse_units(units: Union[str, Unit, None]) -> Union[str, Unit, None]:
+def _parse_units(units: str) -> Unit:
```
kroenlein, line 163: Method with clean signature for internal units conversion.

tests/units/test_parser.py
```diff
@@ -23,22 +22,29 @@ def test_parse_expected():
...
         "g / -+-25e-1 m",  # Weird but fine
-        "ug / - -250 mL"  # Spaces between unaries is acceptable
+        "ug / - -250 mL",  # Spaces between unaries is acceptable
+        "1 / 10**5 degC",  # Spaces between unaries is acceptable
+        "m ** - 1"  # Pint < 0.21 throws DefinitionSyntaxError
     ]
     for unit in expected:
-        parse_units(unit)
+        parsed = parse_units(unit, return_unit=return_unit)
+        assert parsed == parse_units(parsed, return_unit=return_unit)
```
kroenlein, line 37: Ensures that transformations are idempotent.

gemd/units/impl.py
```diff
@@ -3,70 +3,132 @@
...
-        # There's probably something to be said for stashing these, but this sin
-        # should be ameliorated by the LRU cache
-        regex = rf"(?<!=[-+0-9.]){re.escape(substr)}(?!=[0-9.])"
-        valid = "_" + value.replace(".", "_").replace("+", "").replace("-", "_")
-        trailing = "/" if division else ""
-        _REGISTRY.define(f"{valid} = {value} = {value}")
-        input_string = re.sub(regex, valid + trailing, input_string)
+    for scaled_term, number_string, unit_string in todo:
+        regex = rf"(?<![-+0-9.]){re.escape(scaled_term)}(?![0-9.])"
+        stripped = re.sub(r"--", "", re.sub(r"[+\s]+", "", scaled_term))
```
btreecat, line 104: I think this would be cleaner with str.translate rather than re.sub.

gemd/units/impl.py
```diff
@@ -3,70 +3,132 @@
...
+
+
 def _scaling_preprocessor(input_string: str) -> str:
     """Preprocessor that turns scaling factors into non-dimensional units."""
+    blocks: List[List[Token]] = [[]]
+    operator_stack = []
+    for token in tokenizer(input_string):
+        exponent_context = any(t.string in {"**", "^"} for t in operator_stack)
+        if token.type == OP:
+            if token.string not in _ALLOWED_OPERATORS:
```
btreecat, line 42: I think we could make this complex block a function to make this section easier to read

tests/units/test_parser.py
```diff
@@ -111,6 +138,19 @@ def test_conversion():
...
+    # Verify that convert_units respects scaling factors
+    assert -1e-8 < convert_units(100, 'g / 100 mL', 'g/cc') - 1 < 1e-8
+    assert -1e-8 < convert_units(1, "g / 2.5 cm", "g / 25 mm") - 1 < 1e-8
+
+
+def test_get_base_units():
+    """Test that base units & conversions make sense."""
+    from gemd.units.impl import _REGISTRY
+    assert get_base_units("degC") == (_REGISTRY("kelvin"), 1, 273.15)
+    assert get_base_units("degC") == get_base_units(_REGISTRY("degC"))
```
btreecat, line 150: If both functions get_base_units() return the wrong value, the assert is still true. The value being compared, should not rely on a method we are asserting against.

kroenlein: We're verifying consistent polymorphic behavior, i.e. gives the same result for a string and a Unit object.  Though I did make an oversight here, since it should be `_REGISTRY("degC").u`

gemd/units/impl.py
```diff
@@ -112,24 +174,60 @@ def parse_units(units: Union[str, Unit, None]) -> Union[str, Unit, None]:
...
+    [Union[str, Unit, None]]
+        The representation; note that the same type that was passed is returned
+
     """
     if units is None:
-        return None
-    elif units == '':
-        return 'dimensionless'
+        if return_unit:
+            return _REGISTRY("").u
```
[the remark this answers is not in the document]
kroenlein: UnitRegistry has a `__call__` method that's mapped to the `parse_expression` method.  Ultimately, we'll be migrating all these calls to `UnitRegistry.parse_units`, but pint 0.18 has a bug where repeat units () are interpreted incorrectly (e.g. `m*meter` => `Unit(m)`).  Once Python 3.7 is EoL in June, we'll migrate this.

gemd/units/impl.py
kroenlein: Hopefully, this routine goes away and we can just use `UnitRegistry.parse_units` once we stop supporting python 3.7.

gemd/units/impl.py
```diff
@@ -169,4 +291,10 @@ def change_definitions_file(filename: str = None):
     convert_units.cache_clear()  # Units will change
     if filename is None:
         filename = DEFAULT_FILE
-    _REGISTRY = UnitRegistry(filename=filename, preprocessors=[_scaling_preprocessor])
+    _REGISTRY = UnitRegistry(filename=filename,
```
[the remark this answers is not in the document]
kroenlein: Private module global, yes.  Needs to be a variable since it changes if we tell it to load a different units file.

gemd/units/impl.py
kroenlein: I don't think that buys us much unless it's compiled outside the function.

anoto-moniz: Right. Any reason it can't be moved outside the function?

gemd/units/impl.py
```diff
@@ -1,72 +1,174 @@
...
+    for token in token_stream:
+        exponent_context = any(t.string in {"**", "^"} for t in operator_stack)
+        if token.type == OP:
+            if token.string not in _ALLOWED_OPERATORS:
+                raise UndefinedUnitError(f"Unrecognized operator: {token.string}")
+
+            if exponent_context or token.string in {"**", "^", ".", "-", "+"}:
+                # Exponents & unaries do not change context
+                result[-1].append(token)
+            elif token.string not in {}:
```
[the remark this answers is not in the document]
kroenlein: Yeah, that should be an else.

gemd/units/impl.py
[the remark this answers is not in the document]
btreecat: We could build in the scaffolding to enable dynamic reloading of the pint library and it's units file, but rn that would require restarting the service regardless.

anoto-moniz (approved):

PR #191, opened 2023-06-05
Update testing strategy

kroenlein:
This PR has no changes to library functionality.

In this PR, we've 
* updated some library dependencies to better support Python 3.11
* modified the install pattern to automatically test against latest Pint and Pandas
* prevented building libraries (in particular Pandas) during testing to keep things speedy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,12 @@ python:
 - '3.10'
 - '3.11'
 env:
-  - UPGRADES="pint --upgrade pandas --upgrade"
+  - UPGRADES="-U pint pandas"
   - UPGRADES=""
 install:
-  - pip install --only-binary ':all:' -r requirements.txt -r test_requirements.txt -e . $UPGRADES
+  - pip install --only-binary ':all:' -r requirements.txt
+  - pip install --only-binary ':all:' -r test_requirements.txt
+  - pip install $UPGRADES -e .
 script:
 - bash scripts/run_tests.sh
 - cd docs; make html; cd ..;


.travis.yml
```diff
@@ -7,17 +7,10 @@ python:
 - '3.10'
 - '3.11'
 env:
-- PINT_VERSION=0.18
-- PINT_VERSION=0.21
-jobs:
-  exclude:
-    - python: 3.7
-      env: PINT_VERSION=0.21
+  - UPGRADES="pint --upgrade pandas --upgrade"
```
btreecat, line 10: `--upgrade` assumes the package is already installed. I don't think you need to flag those packages for upgrade and the `--upgrade` flag should come after install but before the package and would only be needed once IIRC. 

Example call: https://pip.pypa.io/en/stable/installation/#upgrading-pip

btreecat (changes requested):
I think we can clean up the pip install command, it might need to be broken out into two commands if you want to install a minimum, then force upgrade to a later version.

.travis.yml
```diff
@@ -7,17 +7,12 @@ python:
...
 - '3.11'
 env:
-- PINT_VERSION=0.18
-- PINT_VERSION=0.21
-jobs:
-  exclude:
-    - python: 3.7
-      env: PINT_VERSION=0.21
+  - UPGRADES="-U pint pandas"
+  - UPGRADES=""
```
btreecat, line 11: Is this reassigning `UPGRADES` to the empty string?

btreecat (changes requested):

.travis.yml
kroenlein: No, this is saying run once with the upgrades, and once using the versions specified in the requirements files.

https://docs.travis-ci.com/user/languages/python/#testing-against-multiple-versions-of-dependencies-eg-django-or-flask

btreecat (approved):

.travis.yml
btreecat: TIL! Thanks for the new knowledge :smile:

PR #192, opened 2023-06-05
Augment base enumeration class to behave well in string contexts

kroenlein:
This PR modifies the Base Enumeration class to behave better in string contexts, as well as support synonyms and better string matching.

This does not change any behavior that was previously under testing nor change any previously defined interfaces.

### DIFF
--- a/gemd/entity/attribute/base_attribute.py
+++ b/gemd/entity/attribute/base_attribute.py
@@ -49,7 +49,7 @@ class BaseAttribute(DictSerializable):
         self._value = None
         self._template = None
         self._origin = None
-        self._file_links = None,
+        self._file_links = None
 
         self.value = value
         self.template = template
@@ -106,7 +106,7 @@ class BaseAttribute(DictSerializable):
         """Get the expected type of template for this object (property of child)."""
 
     @property
-    def origin(self) -> str:
+    def origin(self) -> Origin:
         """Get origin."""
         return self._origin
 
@@ -114,7 +114,7 @@ class BaseAttribute(DictSerializable):
     def origin(self, origin: Union[Origin, str]):
         if origin is None:
             raise ValueError("origin must be specified (but may be `unknown`)")
-        self._origin = Origin.get_value(origin)
+        self._origin = Origin.from_str(origin, exception=True)
 
     @property
     def file_links(self) -> List[FileLink]:
--- a/gemd/entity/setters.py
+++ b/gemd/entity/setters.py
@@ -6,7 +6,7 @@ from typing import Union, Iterable, Optional, Callable, Type, TypeVar
 T = TypeVar('T')
 
 
-def validate_list(obj: Union[Iterable[T], T],
+def validate_list(obj: Optional[Union[Iterable[T], T]],
                   typ: Union[Iterable[Type], Type],
                   *,
                   trigger: Callable[[T], Optional[T]] = None) -> ValidList:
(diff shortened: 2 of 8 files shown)


StevenKauwe:
Can I get some context how this is being used? I'm curious why we need to be able to handle strings with mixed lower/uppercase. Where are these strings coming from?

kroenlein:
> Can I get some context how this is being used? I'm curious why we need to be able to handle strings with mixed lower/uppercase. Where are these strings coming from?

If you look at routes supported by citrine-python, you'll see several ostensibly enumerated field entries that are only being represented as strings.  You'll also note the casing choice on those enumerations is actually end-point specific.  This removes sensitivity to a user knowing the casing details of particular routes, through giving access to a Enum that still behaves cleanly in a string context, case-insensitive object deserialization, and easing library interactions.

anoto-moniz (approved):
Seems reasonable to me. I presume you have some place in mind you'd like to use this?

kroenlein:
> Seems reasonable to me. I presume you have some place in mind you'd like to use this?

Every instance of citrine-python's use of BaseEnumeration.  Plus they're just more pleasant.  They've been in gemd-ingest for a long time.

StevenKauwe:
Can you clarify the asserts? I'm confused by line 61 especially.

kroenlein:
> Can you clarify the asserts? I'm confused by line 61 especially.

`       ONE = "One", "1"`
...
`    assert TestEnum.ONE != "1", "Equality worked for synonym"`

The synonym mechanism allows you to have multiple inputs that map to the same enumerated value, but equality doesn't work because the string form is always the first entry: `str(assert TestEnum.ONE) == "One"`.  Many inputs, one output.

gemd/entity/attribute/base_attribute.py
```diff
@@ -49,7 +49,7 @@ def __init__(self,
         self._value = None
         self._template = None
         self._origin = None
-        self._file_links = None,
+        self._file_links = None
```
kroenlein, line 52: Collateral discovery.  Just a cosmetic issue b/c the attribute is initialized 4 lines later.

gemd/entity/attribute/base_attribute.py
```diff
@@ -106,15 +106,15 @@ def _template_type() -> Type:
         """Get the expected type of template for this object (property of child)."""
 
     @property
-    def origin(self) -> str:
+    def origin(self) -> Origin:
```
kroenlein, line 109: Origin is a str, so this is non-breaking and more intuitive.

gemd/entity/setters.py
```diff
@@ -6,7 +6,7 @@
 T = TypeVar('T')
 
 
-def validate_list(obj: Union[Iterable[T], T],
+def validate_list(obj: Optional[Union[Iterable[T], T]],
```
kroenlein, line 9: Collateral discovery.  It accepts a None, initializing to empty list, so why is that not documented?

tests/json/test_json.py
```diff
@@ -112,7 +112,8 @@ def test_enumeration_serde():
     """An enumeration should get serialized as a string."""
     condition = Condition(name="A condition", notes=Origin.UNKNOWN)
     copy_condition = GEMDJson().copy(condition)
-    assert copy_condition.notes == Origin.get_value(condition.notes)
+    assert copy_condition.notes == Origin.UNKNOWN.value
+    assert not isinstance(copy_condition.notes, Origin)
```
kroenlein, line 116: get_value is deprecated, plus this is a better test anyway.

StevenKauwe (approved):
LGTM! The added doc strings are very helpful ❤️.

PR #195, opened 2023-09-08
Migrate away from pkg-resources

kroenlein:
As of Python 3.7, distributing files within a package are supposed to use [importlib.resources](https://docs.python.org/3.10/library/importlib.html?highlight=importlib%20resources#module-importlib.resources) instead.  This was causing a large number of import warnings.

Ostentibly no changes in functionality, though some of these interfaces deserve rethinking.

### DIFF
--- a/gemd/units/impl.py
+++ b/gemd/units/impl.py
@@ -243,13 +243,14 @@ def change_definitions_file(filename: str = None):
         # TODO: Handle case where user provides a units file but no constants file
         target = Path(filename).expanduser().resolve(strict=True)
 
-    _REGISTRY = UnitRegistry(filename=target,
-                             preprocessors=[
-                                 _space_after_minus_preprocessor,
-                                 _scientific_notation_preprocessor,
-                                 _scaling_preprocessor
-                             ],
-                             autoconvert_offset_to_baseunit=True)
+    # TODO: Pint 0.18 doesn't accept paths; must stringify
+    _REGISTRY = UnitRegistry(filename=str(target),
+                             preprocessors=[_space_after_minus_preprocessor,
+                                            _scientific_notation_preprocessor,
+                                            _scaling_preprocessor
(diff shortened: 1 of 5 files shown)


gemd/units/impl.py
```diff
@@ -9,12 +14,25 @@
...
+_TEMP_DIRECTORIES = []
+
+
+def _deploy_default_files() -> str:
+    """Copy the units & constants file into a temporary directory."""
+    default_dir = TemporaryDirectory()
+    _TEMP_DIRECTORIES.append(default_dir)
+
+    units_path = Path(default_dir.name) / "citrine_en.txt"
+    units_path.write_text(read_text("gemd.units", "citrine_en.txt"))
```
kroenlein, line 27: The change in package behavior required a bunch of rework here because the units file imports the constants file via a relative file link, but importlib refuses to give a path-like object (as it may be unpacking a ZIP).

gemd/units/impl.py
```diff
@@ -243,13 +243,14 @@ def change_definitions_file(filename: str = None):
...
         target = Path(filename).expanduser().resolve(strict=True)
 
-    _REGISTRY = UnitRegistry(filename=target,
-                             preprocessors=[
-                                 _space_after_minus_preprocessor,
-                                 _scientific_notation_preprocessor,
-                                 _scaling_preprocessor
-                             ],
-                             autoconvert_offset_to_baseunit=True)
+    # TODO: Pint 0.18 doesn't accept paths; must stringify
```
bfolie, line 246: 0.18 is an old version, yes? We'll be updating Pint soon-ish?

bfolie (approved):

gemd/units/impl.py
kroenlein: Support for 0.18 gets dropped with dropping Python 3.7.

PR #196, opened 2023-09-16
Drop support for python 3.7; migrate toolz & pint accordingly

kroenlein:
This PR drops support for Python 3.7 and removes it from the test suite.

It also bumps the minimum Pint version to 0.19.  0.19 fixed a bug with how unit strings are parsed but we needed to maintain support for 0.18 because that was the maximum Pint version.

It also drops toolz as a primary dependency, as toolz < 0.12 was issuing deprecation warnings.  The ways it was used in the source code were well handled by list comprehensions, and so it's now only in the test requirements.

### DIFF
--- a/gemd/entity/value/discrete_categorical.py
+++ b/gemd/entity/value/discrete_categorical.py
@@ -1,6 +1,4 @@
 """Discrete distribution across several categories."""
-from toolz import keymap
-
 from gemd.entity.setters import validate_str
 from gemd.entity.value.categorical_value import CategoricalValue
 from gemd.entity.bounds import CategoricalBounds
@@ -41,7 +39,7 @@ class DiscreteCategorical(CategoricalValue, typ="discrete_categorical"):
         elif isinstance(probabilities, dict):
             if abs(sum(probabilities.values()) - 1.0) > 1.0e-9:
                 raise ValueError("probabilities must sum to 1.0")
-            self._probabilities = keymap(validate_str, probabilities)
+            self._probabilities = {validate_str(k): v for k, v in probabilities.items()}
         else:
             raise TypeError("probabilities must be dict or single value")
 
--- a/gemd/units/impl.py
+++ b/gemd/units/impl.py
@@ -166,8 +166,8 @@ def _scaling_store_and_mangle(input_string: str, todo: List[Tuple[str, str, str]
 
         if unit_string is not None:
             stripped_unit = re.sub(r"[+\s]+", "", unit_string).replace("--", "")
-            long_unit = f"{_REGISTRY(stripped_unit).u}"
-            short_unit = f"{_REGISTRY(stripped_unit).u:~}"
+            long_unit = f"{_REGISTRY.parse_units(stripped_unit)}"
+            short_unit = f"{_REGISTRY.parse_units(stripped_unit):~}"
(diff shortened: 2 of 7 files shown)


gemd/entity/value/discrete_categorical.py
```diff
@@ -41,7 +39,7 @@ def probabilities(self, probabilities: dict):
         elif isinstance(probabilities, dict):
             if abs(sum(probabilities.values()) - 1.0) > 1.0e-9:
                 raise ValueError("probabilities must sum to 1.0")
-            self._probabilities = keymap(validate_str, probabilities)
+            self._probabilities = {validate_str(k): v for k, v in probabilities.items()}
```
btreecat, line 42: 👍

gemd/util/impl.py
```diff
@@ -441,8 +439,8 @@ def recursive_foreach(obj: Union[Iterable, BaseEntity, DictSerializable],
             func(this)
 
         if cached_isinstance(this, Mapping):
-            for x in concatv(this.keys(), this.values()):
-                queue.append(x)
+            queue.extend(this.keys())
```
btreecat, line 442: 👍

btreecat (dismissed):

gemd/units/impl.py
```diff
@@ -15,18 +16,15 @@
 from pint.errors import UndefinedUnitError, DefinitionSyntaxError  # noqa Import
 
 # Store directories so they don't get auto-cleaned until exit
-_TEMP_DIRECTORIES = []
+_TEMP_DIRECTORY = TemporaryDirectory()
```
kroenlein, line 19: An array was overkill here.

gemd/units/impl.py
```diff
@@ -285,36 +252,6 @@ def _format_clean(unit, registry, **options):
...
-    except AttributeError:  # It was non-dimensional
-        magnitude = parsed
-        result = _REGISTRY("").u
-    if magnitude == 0.0:
-        raise ValueError(f"Unit expression had a zero scaling factor. {units}")
-    if magnitude != 1:
-        raise ValueError(f"Unit expression cannot have a leading scaling factor. {units}")
-    return result
-
-
```
kroenlein, line 317: This is no longer required because dropping 0.18 support drops the buggy code that was causing trouble

gemd/units/impl.py
```diff
@@ -369,7 +306,46 @@ def get_base_units(units: Union[str, Unit]) -> Tuple[Unit, float, float]:
...
+        path = Path(target)
+        os.chdir(path.parent)
+        # Need to re-verify path because of some slippiness around tmp on MacOS
+        _REGISTRY = UnitRegistry(filename=Path.cwd() / path.name,
+                                 preprocessors=[_space_after_minus_preprocessor,
+                                                _scientific_notation_preprocessor,
+                                                _scaling_preprocessor
+                                                ],
+                                 autoconvert_offset_to_baseunit=True
+                                 )
```
kroenlein, line 346: MacOS likes to insert a `private` at the head of the path.

kroenlein:
I'm going to have to force merge since this PR actually breaks 3.7 w/ install, but formally that's not a problem.

btreecat (approved):

PR #204, opened 2024-01-08
Remove documentation build bugs

kroenlein:
This PR primarily serves to remove warnings from the gemd-python documentation build pipeline.  It also brings the gemd-python documentation more in line with the citrine-python docs.  Specifically:
* Moves the version into a separate file, allowing it to be added to the doc build
* Updates stale links
* Corrects formatting and spelling issues in some docstrings
* Adds `__all__` lists to packages people might want import from and reformats the existing lists to be strings.
* Adds Citrine branding

Unfortunately, this does not yet resolve two warnings:
* The automated documentation generation creates a root reference/gemd.rst that is not referenced elsewhere
* Classes that are imported as top-level methods created warnings about ambiguity in cross-reference; however, all generated links correctly link back to the docstrings.


anoto-moniz:
> Unfortunately, this does not yet resolve two warnings:
> 
> * The automated documentation generation creates a root reference/gemd.rst that is not referenced elsewhere
> * Classes that are imported as top-level methods created warnings about ambiguity in cross-reference; however, all generated links correctly link back to the docstrings.

Is there any reason not to fix those warnings? Referencing `reference/gemd.rst` in the main index page is a simple (and useful) fix to bullet one. And even if the generated links all work now, changing the underlying labels to get rid of the warnings seems beneficial, if tedious.

docs/source/conf.py
```diff
@@ -10,9 +10,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
+import gemd
 import os
 import sys
-sys.path.insert(0, os.path.abspath('../../gemd'))
+sys.path.insert(0, os.path.abspath('../..'))
```
anoto-moniz, line 16: Using apidoc causes that `reference/gemd.rst` file to be generated. By having it start outside of the gemd/ folder, you're causing its headline to be whatever folder contains `../..`. So I'd suggest leaving this as `../../gemd`.

docs/source/depth/unit_parsing.rst
```diff
@@ -7,8 +7,13 @@ By default, Pint supports a larger set of units than the Citrine Platform.
...
 
+In support of common patterns in materials science, we permit including scaling factors in a unit of measure.
+For example, industrial researchers may have recorded historical data as ``g / 2.5 cm``.
+While this could be converted a simple SI expression, that would prevent researchers from representing the data
+as originally reported, thus creating a potential source of error during the input process.
+
 Requests for support of additional units can be made by opening an issue in the `gemd-python repository`_ on github.
 
-.. _Pint: https://pint.readthedocs.io/en/0.9/
+.. _Pint: https://pint.readthedocs.io/en/stable/
```
anoto-moniz, line 17: If our docs are gonna point to a dependency's docs, shouldn't they point to the version we currently use? Since they're more likely to not apply if we just always say "stable".

docs/source/index.rst
```diff
@@ -6,19 +6,37 @@
...
+
+.. code::
+
+  pip install gemd
+
+or a specific version can be installed, for example:
+
+.. code::
+
+  pip install gemd==1.0.0
```
anoto-moniz, line 29: Since I imagine many users will just copy this line, let's make it the current version. Of course, it's impractical to update it on every release, but we might as well while we're here.

anoto-moniz (dismissed):
A few little comments, but no blockers.

setup.py
```diff
@@ -1,10 +1,16 @@
 from setuptools import setup, find_packages
+from os import path
 
 packages = find_packages()
-packages.append("")
+
+this_directory = path.abspath(path.dirname(__file__))
+about = {}
+with open(path.join(this_directory, 'gemd', '__version__.py'), 'r') as f:
+    exec(f.read(), about)
```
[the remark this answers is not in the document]
kroenlein: Because I copied this code from citrine-python

https://github.com/CitrineInformatics/citrine-python/blob/54b9fb8a0658dfe0c98315260d1a7e1f972d2a53/setup.py#L10-L11

docs/source/index.rst
kroenlein: Updated.  It's unfortunate that you can't substitute in a code block.

docs/source/depth/unit_parsing.rst
kroenlein: Updated to the version currently in requirements.txt.  Another piece that will require continued attention.

kroenlein:
> > Unfortunately, this does not yet resolve two warnings:
> > 
> > * The automated documentation generation creates a root reference/gemd.rst that is not referenced elsewhere
> > * Classes that are imported as top-level methods created warnings about ambiguity in cross-reference; however, all generated links correctly link back to the docstrings.
> 
> Is there any reason not to fix those warnings? Referencing `reference/gemd.rst` in the main index page is a simple (and useful) fix to bullet one. And even if the generated links all work now, changing the underlying labels to get rid of the warnings seems beneficial, if tedious.

I'd looked at referencing the generated gemd.rst and it really doesn't add anything of quality to the docs landing page.

If you know how to the change the underlying labels, I'm all ears.  I have been unable to find a solution on this.  An example error is:

> WARNING: duplicate object description of gemd.entity.object.ProcessSpec.ingredients, other instance in reference/gemd.entity.object, use :noindex: for one of them

This is because apidoc is crawling the library and saying the method imported into object/__init__.py from object/ProjectSpec.py is ambiguous with the object defined in object/ProjectSpec.py.

setup.py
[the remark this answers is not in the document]
anoto-moniz: Fair enough. Let's change it, unless you have some objection. I'll do the same for the SDK.

anoto-moniz:
> If you know how to the change the underlying labels, I'm all ears. I have been unable to find a solution on this. An example error is:
> 
> > WARNING: duplicate object description of gemd.entity.object.ProcessSpec.ingredients, other instance in reference/gemd.entity.object, use :noindex: for one of them
> 
> This is because apidoc is crawling the library and saying the method imported into object/**init**.py from object/ProjectSpec.py is ambiguous with the object defined in object/ProjectSpec.py.

That's the other ones, where it's a class e.g. "WARNING: duplicate object description of gemd.entity.bounds.categorical_bounds.CategoricalBounds, other instance in reference/gemd, use :noindex: for one of them". The issue is that it's in multiple `__all__` statements, which seems to be a bug in sphinx-apidoc: https://github.com/sphinx-doc/sphinx/issues/11050.

But the one you quote, which is referring to a field, is because the field is given a docstring both in the class docstring and on the field/property itself. The solution is to move that description to the field/property directly, such as I did for citrine-python: https://github.com/CitrineInformatics/citrine-python/pull/910/files#diff-2d687c2d393b3a84762b6673dc4b134d1471c98a87c94bef1e7b004e43cfc4b5

I'd really like to find some solution to that first one, as it results in a long list of warnings, rendering the output of the build much less useful.

setup.py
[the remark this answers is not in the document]
kroenlein: So I think the new version hits your goals, but I'm kinda inclined to just have
```
from gemd.version import __version__
```
even though I know that's not wise.

kroenlein:
> Unfortunately, this does not yet resolve two warnings:
> * The automated documentation generation creates a root reference/gemd.rst that is not referenced elsewhere
> * Classes that are imported as top-level methods created warnings about ambiguity in cross-reference; however, all generated links correctly link back to the docstrings.

Resolves everything but the no-index bug.

gemd/entity/bounds/categorical_bounds.py
```diff
@@ -80,7 +79,7 @@ def union(self,
 
         Parameters
         ----------
-        others: Union[CategoricalBounds, CategoricalValue]
+        others: Union[CategoricalBounds, ~gemd.value.categorical_value.CategoricalValue]
```
anoto-moniz, line 82: Nitpick?
In some places in the docs, you use `Union` when a parameter can accept multiple types. In others, such as `gemd/builders/impl.py` above, you use "or". Is there an intended distinction?

gemd/entity/bounds/composition_bounds.py
```diff
@@ -24,11 +24,12 @@ def __init__(self, components=None):
 
     @property
     def components(self):
```
anoto-moniz, line 26: Nitpick: missing a return type.

gemd/entity/bounds/molecular_structure_bounds.py
```diff
@@ -52,12 +52,12 @@ def union(self,
 
         Parameters
         ----------
-        others: Union[MolecularStructureBounds, MolecularValue]
+        others: Union[MolecularStructureBounds, ~gemd.entity.value.molecular_value.MolecularValue]
             Other bounds or value objects to include.
 
         Returns
         -------
-        CategoricalBounds
```
anoto-moniz, line 58: Wow, these docs were just straight up wrong? Fun!

gemd/entity/has_dependencies.py
```diff
@@ -9,5 +9,8 @@ class HasDependencies(ABC):
     """Mix-in trait for objects that reference other objects."""
 
     @abstractmethod
-    def _local_dependencies(self) -> Set[Union["BaseEntity", "LinkByUID"]]:  # noqa: F821
+    def _local_dependencies(
+            self
+    ) -> Set[Union["gemd.entity.base_entity.BaseEntity",  # noqa: F821
+                   "gemd.entity.link_by_uid.LinkByUID"]]:  # noqa: F821
```
anoto-moniz, line 15: Nitpick: Why does this list subtypes that can be returned? As the parent class, it doesn't know. It only knows (insofar as `typing` can) that each element will be a subclass of `HasDependencies`.

gemd/entity/object/has_conditions.py
```diff
@@ -36,6 +29,9 @@ def conditions(self, conditions: Union[Condition, Iterable[Condition]]):
         checker = self._generate_template_check(HasConditionTemplates.validate_condition)
         self._conditions = validate_list(conditions, Condition, trigger=checker)
 
-    def _local_dependencies(self) -> Set[Union["BaseEntity", "LinkByUID"]]:  # noqa: F821
+    def _local_dependencies(
+            self
+    ) -> Set[Union["gemd.entity.base_entity.BaseEntity",  # noqa: F821
+                   "gemd.entity.link_by_uid.LinkByUID"]]:  # noqa: F821
```
anoto-moniz, line 35: Same as above.

gemd/entity/object/has_material.py
```diff
@@ -22,6 +22,9 @@ def material(self) -> Union[BaseObject, LinkByUID]:
     def material(self, spec: Union[BaseObject, LinkByUID]):
         """Set the material."""
 
-    def _local_dependencies(self) -> Set[Union["BaseEntity", "LinkByUID"]]:  # noqa: F821
+    def _local_dependencies(
+            self
+    ) -> Set[Union["gemd.entity.base_entity.BaseEntity",  # noqa: F821
+                   "gemd.entity.link_by_uid.LinkByUID"]]:  # noqa: F821
```
anoto-moniz, line 28: And again.

gemd/entity/template/has_property_templates.py
```diff
@@ -17,7 +17,7 @@ class HasPropertyTemplates(HasDependencies):
 
     Parameters
     ----------
-    properties: List[(PropertyTemplate, BaseBounds)]
+    properties: List[(~gemd.entity.template.property_template.PropertyTemplate, BaseBounds)]
```
anoto-moniz, line 20: This type for `properties` is quite different from the type hints on the actual parameter.

gemd/entity/template/has_property_templates.py
```diff
@@ -37,7 +37,7 @@ def properties(self) -> List[Tuple[Union[PropertyTemplate, LinkByUID],
 
         Returns
         -------
-        List[(PropertyTemplate, bounds)]
+        List[(~gemd.entity.template.property_template.PropertyTemplate, BaseBounds)]
```
anoto-moniz, line 40: The discrepancies between type hint and docs seem to be present throughout much of this file.

setup.py
```diff
@@ -1,16 +1,23 @@
...
-with open(path.join(this_directory, 'gemd', '__version__.py'), 'r') as f:
-    exec(f.read(), about)
+version_file = path.join(this_directory, 'gemd', '__version__.py')
+version_re = r'''^__version__\s*=\s*(['"])([\w\.]+)\1$'''
+with open(version_file, 'r') as f:
+    mo = re.search(version_re, f.read(), re.M)
+    if mo:
+        version = Version(mo.group(2))
+    else:
+        raise RuntimeError(f"Unable to find version string in {version_file}")
```
anoto-moniz, line 16: Great! Since you got to this before me, I'll adapt your code to citrine-python.

setup.py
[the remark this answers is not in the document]
anoto-moniz: I'm just always wary of an unnecessary `exec`, although it's not the most harmful here, sure. You could use the line you put above. But my personal preference is opening the `__version__` file as plaintext, grabbing only its first line, and regexing for the version string. That is, what you've put in, except I don't think we need to have `__version__` as a variable in the version string file.

But it's not a huge deal either way, so I'll leave the final call for gemd up to you.

gemd/entity/template/has_property_templates.py
kroenlein: An English language description of the type here is a list-like object, where each element of the list is one of:
* A PropertyTemplate
* A LinkByUID that maps to a PropertyTemplate (unknowable / unenforceable)
* A Tuple that contains:
    * A PropertyTemplate or a LinkByUID
    * A Bounds object

The signature lines up with that, but that's a pretty complicated thing to put into documentation.  Probably just needs an essay.

setup.py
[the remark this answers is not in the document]
kroenlein: As weird as it is, as far as I can tell having `__version__` in file is actually best practice.  

https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package

A regex extraction seems gross to me, except it also seems to be the best choice.  We could also try

https://pypi.org/project/setuptools-scm/

if we want to try going down the auto-semantic version approach.

For today, I'll just make this approach work.

gemd/entity/has_dependencies.py
kroenlein: It's not a subclass, it's a mixin / interface specifically designed to say what objects need to be included in a serialization of this object.  It was added to support the register_all method for GEMD objects creating batches of self-supporting objects

https://github.com/CitrineInformatics/citrine-python/blob/54b9fb8a0658dfe0c98315260d1a7e1f972d2a53/src/citrine/_utils/batcher.py#L58

Might make sense to make it an abstract method on BaseEntity instead, except it's used by BaseObject & BaseTemplate & BaseAttribute objects, but not by other Entities.  The object model is really not clean in its dependencies.

gemd/entity/object/material_run.py
```diff
@@ -93,8 +86,15 @@ def process(self, process: Union[ProcessRun, LinkByUID]):
             raise TypeError("process must be a ProcessRun or LinkByUID: {}".format(process))
 
     @property
-    def measurements(self) -> List["MeasurementRun"]:  # noqa: F821
-        """Get a read-only list of the measurement runs."""
+    def measurements(
+            self
+    ) -> List["gemd.entity.object.measurement_run.MeasurementRun"]:  # noqa: F821
```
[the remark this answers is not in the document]
anoto-moniz: Ken clarified in Slack this is due to circular dependencies.

gemd/entity/template/has_property_templates.py
anoto-moniz: If we're going to document the types, then we should be as accurate and precise as possible. Otherwise, there's no point.

If we can't describe the types we accept as English, then using a code like structure similar to the type hint is a reasonable move. Because while it's not as important that the documentation of inputs cover every case, that both type hints show up in the docs may cause some confusion.

But I'm more concerned with disagreements between the docstrings and type hints with properties and other values to be read. For example, `parameters` documents its return as `List[Union[ParameterTemplate, LinkByUID]]`, which is quite different from the docstring's claim of ` List[(ParameterTemplate, BaseBounds)]`. A list of elements which are either `ParameterTemplate` or `LinkByUID` objects is not the same as a list of tuples, where each tuple is one `ParameterTemplate` followed by a `BaseBounds`.

Although this raises the question of why we must include the type in the docstring and inline. It would seem to be redundant, as the generated Sphinx docs include both, and thus just introduces chances for conflict, like this.

gemd/entity/has_dependencies.py
kroenlein: A valid return value from _local_dependencies is a PropertyTemplate, because something can depend on a PropertyTemplate.  A PropertyTemplate is not a HasDependencies, because it can't depend on anything.  

The return values from _local_dependencies are not necessarily HasDependencies objects.

anoto-moniz (approved):
I've registered my disagreements on the way we're documenting types here. They're not accurate, and bound to be confusing to users should they look at the documentation.

However, this is just documentation, not a bug. Worst come to worst, they'll learn the discrepancies when they go to use it. More importantly, these are not discrepancies you introduced: they already existed, and you just happened to touch the lines bringing them to my attention.

As such, I'll step out of the way, since the PR doesn't make anything worse, and improves a bunch of stuff. In the future, I would like to see us be better about documenting types and keeping docstrings in line with type hints if we're to continue using them.

kroenlein:
This should address everything but the inconsistencies in types for the has_X_templates family of modules.  We can revisit, but I think part of the challenge is that the setters are very forgiving, and the `Parameters` fields line up with the returns (or at least I think they do) not with the acceptable incoming data.

test_requirements.txt
```diff
@@ -5,6 +5,6 @@ pytest-cov==4.0.0
 pandas==1.5.0
 toolz==0.12.0
 derp==0.1.1
-sphinx==7.1.2
-sphinx-rtd-theme==2.0.0
-sphinxcontrib-apidoc==0.5.0
+sphinx==5.0.0
+sphinx-rtd-theme==1.0.0
+sphinxcontrib-apidoc==0.3.0
```
anoto-moniz, line 10: What's the reason for reverting these requirements?

anoto-moniz (approved):

PR #206, opened 2024-01-10
Move real/integer bounds attributes to properties

kroenlein:
This PR migrates the lower_bound and upper_bound attributes of IntegerBounds and RealBounds objects to properties, creating getters & setters and locating validation code in those subroutines.  It also moves responsibility for verifying that required positional arguments were passed from an explicit default argument check to the compiler.

Technically breaking, since it migrates what was formerly a ValueError to a TypeError and adds setter constraints that were previously bypassed if the user modified objects.  Such objects were already considered invalid by Citrine Platform.

This should represent no meaningful change in behavior.

### DIFF
--- a/gemd/entity/bounds/integer_bounds.py
+++ b/gemd/entity/bounds/integer_bounds.py
@@ -10,7 +10,7 @@ __all__ = ["IntegerBounds"]
 class IntegerBounds(BaseBounds, typ="integer_bounds"):
     """Bounded subset of the integers, parameterized by a lower and upper bound."""
 
-    def __init__(self, lower_bound, upper_bound):
+    def __init__(self, lower_bound: int, upper_bound: int):
         self._lower_bound = None
         self._upper_bound = None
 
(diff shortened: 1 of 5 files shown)


gemd/entity/bounds/integer_bounds.py
```diff
@@ -1,34 +1,51 @@
...
 
-    Parameters
-    ----------
-    lower_bound: int
-        Lower endpoint.
-    upper_bound: int
-        Upper endpoint.
+    def __init__(self, lower_bound, upper_bound):
+        self._lower_bound = None
+        self._upper_bound = None
```
anoto-moniz, line 15: These should be assigned to the parameters to preserve backwards compatible behavior.

anoto-moniz: Oops, missed the lines below. Nevermind.

gemd/entity/bounds/integer_bounds.py
```diff
@@ -1,34 +1,51 @@
...
+    def upper_bound(self) -> int:
+        """The upper endpoint of the permitted range."""
+        return self._upper_bound
+
+    @upper_bound.setter
+    def upper_bound(self, value: int):
+        """Set the upper endpoint of the permitted range."""
+        if value is None or not isfinite(value):
+            raise ValueError(f"Upper bound must be given and finite: {value}")
+        if value < self.lower_bound:
```
anoto-moniz, line 45: Why not just do the `self.lower_bound is not None` check here, too, so they can be set in any order? I know we're the only ones who'll be setting it when their values are `None`, but seems like unnecessary enforced ordering.

anoto-moniz (dismissed):
Only comment is the one about unnecessary ordering, but it only impacts us, it's not a big deal if you wanna ignore it.

anoto-moniz (approved):

PR #208, opened 2024-02-01
Maintain/2 0 remove deprecations

kroenlein:
This PR removes methods that were deprecated, scheduled for removal w/ the 2.0 release.
* Drop `name` argument from `LinkByUID.from_entity`
* Drop type-casting methods `get_value` and `get_enum` from `BaseEnumeration`
* Drop `register_classes` meta-method from `GemdJson`
* Drop `native_uid` argument from `substitute_links` from util


tests/json/test_json.py
```diff
@@ -1,5 +1,5 @@
 """Test serialization and deserialization of gemd objects."""
-import json
+import json as json_builtin
```
anoto-moniz, line 2: Why? I don't see anything else called `json`, and even if there was, I'd suggest aliasing the other thing rather than the built-in.

anoto-moniz (dismissed):
LGTM. Just that one comment, but it's on a test file and not a big deal.

anoto-moniz (approved):
As noted in Slack, I'm very opposed to the renaming of the builtin `json` import on the basis of clarity, and even more since it's to cater to a specific IDE.

But as I also noted, I don't care enough to argue about it, so I won't block on it.

anoto-moniz (approved):

anoto-moniz (approved):

PR #209, opened 2024-02-02
Prevent count, radian, bit and dimensionless interconversion

kroenlein:
Pint will convert between various expressions of non-dimensional numbers, even when the base derived units aren't quite compatible.  This puts guardrails around this, throwing the conversion-error exception as normal conversion issues.


anoto-moniz (approved):
Seems reasonable to me.

PR #210, opened 2024-02-02
Maintain/boost test requirements versions

kroenlein:
Library clean-up

* Update library versions for testing libraries
* Use a range of acceptable pandas installs for testing, to remove need for conditioning test_requirements.txt
* Remove [toolz](https://pypi.org/project/toolz/) dependency, since convenience functions were test-facing only
* Remove unused scripts/validate-version-bump.sh
* Pull doc building installs into an additional file, to improve build speed


anoto-moniz (approved):
lgtm

PR #212, opened 2024-04-02
Add custom UnitRegistry to handle scaling factors more cleanly

kroenlein:
An issue happened where stringified versions of parsed units weren't directly parsable on later invocations of the library when there was a scaling factor.  For example, `grams / 10 minutes` was being internally represented as `gram / _10_minute`, but then the units service was storing that instead of what would have come out of a `clean` formatted string.  The specific bug encountered is here:

https://citrine.atlassian.net/browse/SPT-1311

This PR prevents this issue in the future by:
* Subclassing Unit so that it's `__format__` method always cleans up the munged scaling factor
* Deprecating `clean` since the subclass makes it obsolete
* Adding testing


gemd/units/impl.py
```diff
@@ -201,7 +207,58 @@ def _scaling_preprocessor(input_string: str) -> str:
     return _scaling_store_and_mangle(input_string, todo)
 
 
-_REGISTRY: UnitRegistry = None  # global requires it be defined in this scope
+def _unmangle_scaling(input_string: str) -> str:
```
btreecat, line 210: :laughing:

gemd/units/impl.py
```diff
@@ -201,7 +207,58 @@ def _scaling_preprocessor(input_string: str) -> str:
...
+    # https://pint.readthedocs.io/en/0.23/advanced/custom-registry-class.html
+    from pint.registry import GenericUnitRegistry
+    from typing_extensions import TypeAlias
+
+    class _ScaleFactorUnit(UnitRegistry.Unit):
+        """Child class of Units for generating units w/ clean scaling factors."""
+
+        def __format__(self, format_spec):
+            result = super().__format__(format_spec)
+            return _unmangle_scaling(result)
```
robbtraister, line 234: this part is the same in both cases, right? could it be lifted up out of the try block to avoid duplication?

tests/units/test_parser.py
```diff
@@ -205,8 +209,17 @@ def test_exponents():
...
-    assert "1e2 kg" in parse_units("F* 10 ** 2 kg")
+    assert "1e2 kilogram" in parse_units("F* 10 ** 2 kg")
+    assert "1e2 kg" in f'{parse_units("F* 10 ** 2 kg", return_unit=True):~}'
     assert "1e-5" in parse_units("F* mm*10**-5")
     assert "1e" not in parse_units("F* kg * 10 cm")
     assert "-3.07e2" in parse_units("F* -3.07 * 10 ** 2")
     assert "11e2" in parse_units("F* 11*10^2")
+
+
+def test_deprecation():
```
btreecat, line 220: Best practice assume 3p libs are working

btreecat (approved):

gemd/units/impl.py
```diff
@@ -244,38 +301,23 @@ def convert_units(value: float, starting_unit: str, final_unit: str) -> float:
...
+    Responsibility for this piece of clean-up has been shifted to a custom class.
+
+    """
+    from pint.formatting import _FORMATTERS
+    return _FORMATTERS["D"](unit, registry, **options)
 
 
 @functools.lru_cache(maxsize=1024)
-def parse_units(units: Union[str, Unit, None],
+def parse_units(units: Union[str, _ScaleFactorUnit, None],
```
robbtraister, line 317: does this mean I can't pass in a base `Unit` instance?

gemd/units/impl.py
kroenlein: Yeah, I thought of that and then forgot... 🤦   Thanks for the reminder.

gemd/units/impl.py
kroenlein: It'd work, but typehints would complain.  Functionally there's no reason to not support both, so I'll update that.

gemd/units/impl.py
kroenlein: I tried abstracting one layer further, but it broke the magic methods.  The `_unmangle_scaling` is abstracted, so that's probably DRY enough.

gemd/units/impl.py
```diff
@@ -244,38 +301,23 @@ def convert_units(value: float, starting_unit: str, final_unit: str) -> float:
 
 
 @register_unit_format("clean")
+@deprecated(deprecated_in="2.1.0", removed_in="3.0.0", details="Scaling factor clean-up ")
```
robbtraister, line 304: if this is deprecated, what is it replaced with?

robbtraister: is it that `_unmangle_scaling` encompasses the customization that clean was performing before?

kroenlein: Yes, the clean formatter is now obsolete, because of the `__format__` method does its fundamental work without getting in the way of invoking other formatters (`H`, `P`, etc.).  That's potentially very valuable because we'll because to do server-side HTML formatting, rather than forcing a browser side fix.

PR #213, opened 2024-04-03
Update resource import strategy to resolve deprecation warnings

kroenlein:
importlib has deprecated several methods in favor of methods that are not available in Python 3.8.  This PR imports their third-party re-implementations to maintain 3.8 support while resolving the warnings for 3.12 in this and downstream libraries.


gemd/demo/cake.py
```diff
@@ -67,10 +68,7 @@ def get_template_scope():
 
 def import_toothpick_picture() -> BytesIO:
     """Return the stream of the toothpick picture."""
-    from importlib.resources import read_binary
-    resource = read_binary("gemd.demo", "toothpick.jpg")
-
-    return BytesIO(resource)
+    return files("gemd.demo").joinpath("toothpick.jpg").open("rb")
```
mVenetos97, line 71: Should we rewrite this to `files("gemd.demo").joinpath("toothpick.jpg").read_bytes()` to be consistent with `strehlow_and_cook.py` and `test_parser.py`?

kroenlein: It'd be nice, but that would break downstream consumers that expect a byte stream instead of bytes.

mVenetos97 (approved):
Looks good to me

PR #214, opened 2024-04-17
Migrate to github actions

kroenlein:
Migrate gemd-python to use Github Actions instead of travis + jenkins.

* Add actions files
* Modify Sphinx install to be in install action, not doc building script
* Remove packaging as a dependency of setup.py, since it's not core


.github/workflows/pr-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+      - 'release/**'
+
+jobs:
+  run-tests:
+    name: Execute unit tests
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
anoto-moniz, line 16: Side note: we should probably add 3.12 support to the SDK. I'll cut a ticket, and do it next time I'm thinking about the SDK, probably next week.

anoto-moniz (approved):

PR #217, opened 2024-05-14
Migrate to common-gh-actions

kroenlein:
# GEMD Python PR

## Description 
This PR moves to using the common-gh-actions repo instead of locally defined actions.

### PR Type:
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Maintenance (non-breaking change to assist developers)

### Adherence to team decisions
- [ ] I have added tests for 100% coverage
- [ ] I have written Numpy-style docstrings for every method and class.
- [x] I have communicated the downstream consequences of the PR to others.
- [x] I have bumped the version in __version\__.py


kroenlein:
Reverted update to deploy.yml as per https://github.com/CitrineInformatics/common-gh-actions/pull/3

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,21 @@
...
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
```
anoto-moniz, line 17: I found that this stanza actually causes the workflow to be cancelled, since it also exists in the common actions version. I think best is to leave it in the common actions workflow, so each caller doesn't have to worry about it.

kroenlein: Sounds reasonable. 👍

anoto-moniz (changes requested):
Just that one issue. Other than that, looks good!

anoto-moniz (approved):

PR #225, opened 2025-06-27
Drop 3.8 support; add 3.13 testing; drop support for old versions of Pint

kroenlein:
# GEMD Python PR

## Description 
This PR:
* Removes official support & testing for Python 3.8, though no code is broken by this change in dependencies
* Adds testing and compatibility for Python 3.13
* Pins (Pint)[https://github.com/hgrecco/pint] to version 0.24.4, since older versions interact in some complex and confusing ways with numpy because of the migration of the cumproduct method.

### PR Type:
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Maintenance (non-breaking change to assist developers)

### Adherence to team decisions
- [ ] I have added tests for 100% coverage
- [ ] I have written Numpy-style docstrings for every method and class.
- [x] I have communicated the downstream consequences of the PR to others.
- [x] I have bumped the version in __version\__.py


gemd/units/impl.py
```diff
@@ -307,7 +307,7 @@ def _format_clean(unit, registry, **options):
     """
     try:  # Informal route changed in 0.22
         from pint.formatting import _FORMATTERS
-        formatter = _FORMATTERS["D"]
+        formatter = _FORMATTERS["D"]  # pragma: no cover
```
kroenlein, line 310: This is unreachable with the current library limits, but is reachable for older versions of Pint.  Kept to avoid being "breaking" without a major version bump.

test_requirements.txt
```diff
@@ -1,6 +1,9 @@
 flake8==7.0.0
 flake8-docstrings==1.7.0
+numpy==1.24.4; python_version<'3.10'
+pandas==2.0.3; python_version<'3.10'
+numpy>=2.0.2,<=2.1.0; python_version>='3.10'
+pandas==2.3.0; python_version>='3.10'
```
kroenlein, line 6: There are some unfortunate import errors if we're not careful about numpy/pandas/Pint combinations

StevenKauwe (approved):
lgtm

